PR #2, opened 2021-02-05
Week4 - Project Chatbot

skossan:
We struggled to get the event, preventDefault to work so we had to use an "onsubmit="return false;" on our form which made our code work but we don't think this is the right way to do it.

We used innerHTML to remove and add html code. We suspect that there might be better ways of dealing with this. We ran into several issues with innerHTML, where we accidentally overwrote certaing things we didn't mean to touch at all. InnerHTML felt a bit tedious to use, it was difficult to forsee what the result would be/look like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,16 @@
 # Project Name
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+The project objective was to work in pairs and build a chatbot.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+The goal was to practice our conditional as well as function skills using vanilla javascript. We stuck to basic use of variables, functions and if-statements.
+
+We encountered several bugs relating to innerHTML that took us some time to wrap our heads around but in the end we managed to figure out a way to solve them.
+
+If we had more time we would for example have done some styling. We would also try other forms of inputs, such as radio buttons, images etc, regular buttons, etc.
 
 ## View it live
 
-Have you deployed your project somewhere? Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+Project can be found here:
+https://tender-dijkstra-73ca91.netlify.app/
--- a/code/script.js
+++ b/code/script.js
@@ -2,6 +2,11 @@
 const chat = document.getElementById('chat')
 const formSubmitButton = document.getElementById("send-btn");
 const formMessageBox = document.getElementById("form-input");
+const formInputWrapper = document.getElementById("input-wrapper");
+const selectMenuOptions = document.getElementById("selectMenuOptions");
+const selectOptionButton = document.getElementById("selectOptionButton");
+const selectOptionDiv = document.getElementById("selectOptionDiv");
+const mainTag = document.getElementById("mainTag");
 
 // Global variables, if you need any, declared here
 let questionNumber = 1;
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -1,13 +1,16 @@
...
+
+We encountered several bugs relating to innerHTML that took us some time to wrap our heads around but in the end we managed to figure out a way to solve them.
+
+If we had more time we would for example have done some styling. We would also try other forms of inputs, such as radio buttons, images etc, regular buttons, etc.
 
 ## View it live
 
-Have you deployed your project somewhere? Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+Project can be found here:
+https://tender-dijkstra-73ca91.netlify.app/
```
KarolCislowski, line 16: README file is a very special place in your repo. You should try to 'sell' your project here. 
Use it to tell your users what your application does, how to use it and eventually what next functionalities can be added in the future. 
Make it shine because this is cover of your project and you need to make your visitors stop for a while and look deeper to your code.

code/script.js
```diff
@@ -1,47 +1,282 @@
...
+const formInputWrapper = document.getElementById("input-wrapper");
+const mainTag = document.getElementById("mainTag");
 
 // Global variables, if you need any, declared here
 
+/* This varaible will keep track of which question functin that will be triggered in the IF statement. */
+let questionNumber = 1;
+
+/* This variable will set a global chat delay to every bubble to emulate a more natural conversation flow. */
+let chatDelay = 1000;
```
KarolCislowski, line 14: You do not change the value, so you should use `const` instead of `let` 😺

code/script.js
```diff
@@ -1,47 +1,282 @@
...
+
+    /* We had to add a new event lsitener to our "buy" button we rendered above. */
+    selectOptionButton.addEventListener("click", () =>
+      nextQuestion(selectMenuOptions.value)
+    );
+  } else {
+    botReply("You have to type either low risk or high risk.");
+  }
+};
+
```
KarolCislowski, line 185: You created 2 very long and very similar functions. I think it is a better way to create a more generic "template" function, and pass data to it via parameters like it is done in `showMessage` function 🤓

code/script.js
```diff
@@ -1,47 +1,282 @@
...
+  } else if (questionNumber === 4) {
+    // Function call
+    highRiskOrLowRiskFundQuestion(message);
+  } else if (questionNumber === 99) {
+    // Function call
+    finalQuestion(message);
+  } else {
+    botReply("questionNumber invalid value!");
+  }
+};
```
KarolCislowski, line 90: Nice way to provide the flow of the application 👍🏻

code/style.css
```diff
@@ -128,13 +129,28 @@ button {
...
   font-size: 16px;
   line-height: 26px;
-  font-family: 'Montserrat';
+  font-family: "Montserrat";
   font-weight: 500;
   cursor: pointer;
   transition: all 0.3s ease;
 }
 
+.selectOptionDiv {
```
KarolCislowski, line 138: According to Technigo code guidelines, CSS selectors should be written in `kebab-case`

KarolCislowski (changes requested):
Your code is very advanced 🥇  You can think for the next time to create more shorter, simpler functions.
I can't wait to see your next projects!
I, unfortunately, couldn't open the netlify link. Please check it, update, and add your names because not all code reviewers are teachers, and know the team's members 😸

skossan:
This was a partner project with myself and Ann-Sofie Jönsson.
New netlify link: https://unruffled-wing-0ab879.netlify.app

KarolCislowski (approved):
Thank you for your fast response 😸

PR #10, opened 2021-02-05
Project - Chatbot, Räkna med Rut

TobiasHagberg:
Hi,

Project - Chatbot, Räkna med Rut

Tobias Hagberg
Elle Frankenberg
Team Tigers - Jan2021


code/script.js
```diff
@@ -19,29 +24,134 @@ const showMessage = (message, sender) => {
...
+      }
+  } else if (step === 11) {
+      if (Number(userInput) === 20) {
+        showMessage(`Hipp hipp hurra va du va bra! Du klarade allt! För att köra igen behöver du ladda om sidan 👋🏼`, "bot");
+        form.innerHTML = "";
+      } else {
+        showMessage(`Det var tyvärr fel, prova igen`, "bot");
+      }  
+  }
+});
```
KarolCislowski, line 139: When you have a long `if/else` it is good to use `switch` statement instead. It is cleaner and easier to read and maintain. You can read about it here: https://www.w3schools.com/js/js_switch.asp

code/script.js
```diff
@@ -19,29 +24,134 @@ const showMessage = (message, sender) => {
...
+          step = 2;
+      }
+  } else if (step === 2) {
+
+      if (Number(userInput) === 2) {
+          showMessage(`Rätt svar! Vad är 2 x 2?`, "bot");
+          step = 3;
+      } else {
+          showMessage(`Det var tyvärr fel, prova igen!`, "bot");
+      }
```
KarolCislowski, line 74: This construction is repeated a lot in your code. It is a good practice to extract code like this and create a function which would be called when needed. 🤓

code/script.js
```diff
@@ -19,29 +24,134 @@ const showMessage = (message, sender) => {
...
+  } else if (step === 3) {
+      if (Number(userInput) === 4) {
```
KarolCislowski, line 76: You can convert string to number by adding `+` on front of the name like this  `+userInput === 4`  😸

code/script.js
```diff
@@ -19,29 +24,134 @@ const showMessage = (message, sender) => {
...
+          showMessage(`Tjohej va duktig du är! Vad är 2 x 6?`, "bot");
+          step = 4
+      } else {
+          showMessage(`Åh nej det var tyvärr fel, testa igen`, "bot");
+      }
+  } else if (step === 4) {
+      if (Number(userInput) === 12) {
```
KarolCislowski, line 83: Isn't it will be better to change your input type to number? This would save you trouble with conversion and saved you from getting non-number values from your users. 🤓

KarolCislowski (approved):
Hi, I'm Karol Cislowski and I'm Technigo's teacher assistant. Sometimes it will be me providing Code Reviews for your assignments. If you have any questions or want to chat about code then feel free to reach me on Technigo's Slack. So let's come back to our business 😸 

This is a solid project. 🦾 
Try to experiment some more with your code! Don't be scared to break it, because repairing is the best way to learn how something works 🔨 Below I provided some tips about what you can do with your code, or apply in the future projects to rise it to the next level 🚀 

Make sure to work more with your styling. It is the only way to master CSS: doing it more and more. 

I can wait to see your next projects!

PR #19, opened 2021-02-06
Project Chat Bot: Hannah and Ylva

jesinkey:
Hannah Jesinkey and Ylva Landoff Lindberg


code/script.js
```diff
@@ -1,47 +1,105 @@
...
+let questionCounter = 0
 
-// Functions declared here
+// Functions declared here.
+const userReply = () => {
+    showMessage(userInput.value, 'user')
+}
+const botReply = (message) => {
+    showMessage(message, 'bot')
+}
```
KarolCislowski: you could add setTimeout here, what would make your code much cleaner and shorter 😸

code/script.js
```diff
@@ -1,47 +1,105 @@
...
+        } else if (userInput.value === `2`) {
+            botReply(`Make sure to create balance in your life. Try to find time to exercise and get some daylight every day, but also time to slow down and recharge your battery. Try to be social but take it easy with alcohol, it messes up your sleep.`)
+        } else if (userInput.value === `3`) {
+            botReply(`That’s amazing!`)
+        } else {
+            botReply(`I am afraid I don't understand you answer, please try again or contact another health professional`)
+        }
+
+        if (userInput.value === `1`) {
+            setTimeout(botReply, 1000, [`So ${name}, I see our time is up. Keep on breathing and I see you next week`])
```
KarolCislowski: I don't understand why you put your strings inside the array here❓

code/script.js
```diff
@@ -1,47 +1,105 @@
...
+            setTimeout(botReply, 1000, [`So ${name}, I see our time is up. Nice talking to you! Maybe we can go out parting some day!?`])
+        }
+
+        setTimeout(botReply, 3000, [`Oh, by the way. I’m sending you the bill to your IP address. Bye bye!`])
+    }
+
+    console.log(userInput.value)
+    userInput.value = '';
+    questionCounter = questionCounter + 1;
+});
```
KarolCislowski: This is a very long function. Please try to cut it on the smaller functions. 🔪

code/style.css
```diff
@@ -1,140 +1,149 @@
...
+    margin: auto 0;
+    background: white;
+}
+
+.intro-text {
+    font-weight: bold;
+    font-size: 24px;
+    line-height: 34px;
+    color: white;
+    ;
```
KarolCislowski: This semicolon can mess with your styling 🤓

KarolCislowski (approved):
Hi, I'm Karol Cislowski and I'm Technigo's teacher assistant. Sometimes it will be me providing Code Reviews for your assignments. If you have any questions or want to chat about code then feel free to reach me on Technigo's Slack. So let's come back to our business 😸 

I can see that you had a lot of fun with it and it is the most important 👍🏻 
Try to experiment some more with your code! Don't be scared to break it, because repairing is the best way to learn how something works 🔨 Below I provided some tips about what you can do with your code, or apply in the future projects to rise it to the next level 🚀 

I can wait to see your next projects!

code/script.js
jesinkey: It was just to make sure that a string was being compared to a string. As I understand it for 3*= it is necessary that the two sides are the same types. And when using the .value is always returns a string, I think?

code/script.js
jesinkey: I don't quite understand?  Can you clarify a bit more?

KarolCislowski:
``` 
const botReply = (message, time) => {
    setTimeout(showMessage(message, 'bot'), time)
}
```
This would save you a lot o typing...

code/script.js
jesinkey: Hi! We had trouble doing this as this was a single conversation and didn't know where to cut it up?

code/script.js
jesinkey: Ah OK! we fixed that in the new push

jesinkey:
Hi! We took your advice on the setTimeout and changed the botReply function. We could not think of how to cut up the eventListener with the conversation as it is one continuos if-string. We added comments for better clarity. Let us know what you think!

PR #43, opened 2021-09-10
Week 4 Chat Bot Amanda & Naushin

amandatilly:
https://vibrant-newton-d20142.netlify.app/

In order to create this chat bot we tried to use the functional programming approach. We created two functions to handle user input and the questions asked by the bot respectively.

If we had more time we would have tried to have more logical conversation between the bot and the user, a better ending and adding more functions instead of repeating code and last but not least some more styling through CSS.


code/script.js
```diff
@@ -1,9 +1,143 @@
...
+let currentQuestion = 1 //tracks the bot questions
+let userInput  //handles user input, button, text, menu
+// DOM selectors for all dynamically created elements are stored in the following variables
+let foodSelected 
+let serving
+let drink
+let selectedDrink
+ 
 // Functions declared here
+//changes color of drop down menu and displays options
```
Meeteyes, line 17: Does that apply to the function that starts on line 19?

code/script.js
```diff
@@ -1,9 +1,143 @@
...
+
+//handles all the user inputs
+const handelInput = (event) => {
+  event.preventDefault()
+  event.stopPropagation()
```
Meeteyes, line 22: Nice that they included Propagation().

code/script.js
```diff
@@ -1,9 +1,143 @@
...
+
+  if (currentQuestion===1) {  //first question
+   userInput = userName.value  //storing the entered text in UserInput variable
+  userName.value = ""  //clearing the text entered by user
```
Meeteyes, line 26: Nice that you clean up the input right away

Meeteyes (commented):
You have done a very good job with the JS file. It took only a couple of minutes to understand what are you doing in each line. The code is well commented, so it was easy to follow.
We thought that you could have saved a couple of lines by combining the two separate conditional blocks into one which would make it easy to read.
Looking at the finished site, maybe adding a button to confirm the selection of the dish size would be beneficial.
You fulfilled all of the blue, red and some black requirements.
Great job!
Reinis, Lena, Jessica.

PR #50, opened 2021-09-11
Hedvigs and Emelies Chatbot

HedvigM:
we are al in al satisfied with the result, even though its a bit "simple" and we had big ambitions with it in the start. Some illness got in the way for the teamwork and we ended up coding and figuring stuff out a bit on our own and a bit together. In the end we merged the code together and made it work. 

As I made the link to Netlify I saw that the sounds that are supposed to come when the different chat bubbles appear, is not in  sync... It worked in the live server, so this problem appeared just as we were to hand the project in. 

If we had more time we would like to maybe add some items to the order and make a "order presentation" in the end, and also to make the sounds sync at all times.


code/index.html
```diff
@@ -9,11 +9,15 @@
       href="https://fonts.googleapis.com/css2?family=Montserrat:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap"
       rel="stylesheet" />
     <title>Chat Bot</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com"> 
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin> 
+<link href="https://fonts.googleapis.com/css2?family=Special+Elite&display=swap" rel="stylesheet">
+<audio id="clickUser" src="assets/click.wav"></audio>
```
isomoth, line 15: This is good to know! We didn't use audio on our bot, so it's good info for the future.

code/script.js
```diff
@@ -1,22 +1,28 @@
...
       <section class="user-msg">
         <div class="bubble user-bubble">
           <p>${message}</p>
         </div>
-        <img src="assets/user.png" alt="User" />  
+        <img src="assets/user.png" alt="User" />
       </section>
     `
   } else if (sender === 'bot') {
+    let sound = new Audio("assets/botSound.wav")
```
isomoth, line 24: Using different sounds for the bot and the user is a great idea in terms of accessibility! People who guide themselves by sound will appreciate this feature a lot :)

code/script.js
```diff
@@ -26,22 +32,117 @@ const showMessage = (message, sender) => {
...
+
+  } else (questionNumber === 4)
+}
+
+
 // Starts here
 const greeting = () => {
-  showMessage(`Hello there, What's your name?`, 'bot')
-  // Just to check it out, change 'bot' to 'user' here 👆
+  questionNumber = 1
```
isomoth, line 59: This is good solution if the number of questions is reduced, like in this case. But we're curious: Did you also consider using `questionNumber++` instead?

code/script.js
```diff
@@ -26,22 +32,117 @@ const showMessage = (message, sender) => {
...
+  } else if (questionNumber === 2) {
+    showMessage(message, 'user') 
+    handleNameInput.value = ''
+    setTimeout(() => drinks(message, 'user'), 1000)
+
+  } else if (questionNumber === 3) {
+    showMessage(message, 'user')
+    setTimeout(() => drinkChoices(message), 1000)
+
+  } else (questionNumber === 4)
```
isomoth, line 53: We were curious at first: What's question nr. 4? We know you define it further down below, but maybe adding a comment in here could also help? Just for extra clarity :)

isomoth (commented):
Your code is really clean, and we didn't have any problems following it. Good job! 

We also didn't find any problems with the sound. We're not sure if the first bot question was supposed to sound as well? It isn't a major detail in any case, and it doesn't affect the user's experience in our opinion. 

Do you have plans to create a README file? It would help a lot, especially because of the Netlify link. We managed just fine by downloading the project and running it in our browser, but it could be a good guide for future projects :)

Keep it up! 

// Nina and Isabel

PR #52, opened 2021-09-12
Pepbot - Isabel & Nina 

NinaAlejandra:


code/index.html
```diff
@@ -1,33 +1,34 @@
...
 
-  <body>
-    <h1>Welcome to xxx.</h1>
-    <h2>Ask about xxxx.</h2>
-    <main>
-      <section class="chat" id="chat"></section>
+<body>
+  <h1>Welcome to Peppbot!</h1>
+  <h2>Looking for a bit of a pick-me-up?</h2>
+  <button id="start-btn">START</button>
```
IdaAspen, line 17: Nice feature with a start button for the bot!

code/script.js
```diff
@@ -1,14 +1,30 @@
...
+// This function adds a chat bubble in the correct place based on who the sender is
 
-// This function will add a chat bubble in the correct place based on who the sender is
 const showMessage = (message, sender) => {
   if (sender === 'user') {
-    chat.innerHTML += `
+
+    // Isabel's console.log (just as a test)
+
+    console.log(message)
```
IdaAspen, line 17: For next time, maybe remove the console.log for the final push

code/script.js
```diff
@@ -26,22 +42,221 @@ const showMessage = (message, sender) => {
...
-  showMessage(`Hello there, What's your name?`, 'bot')
-  // Just to check it out, change 'bot' to 'user' here 👆
+  showMessage(`Hi there! What's your name?`, 'bot')
+}
+
+const moodQuestion = () => {
+  
+  showMessage(`That's a nice name! How are you feeling today?`, 'bot')
+
+  // Generate a set of buttons with mood options
```
IdaAspen, line 58: Perfect with all those comments so the code reviewer easily can understand your code

code/script.js
```diff
@@ -26,22 +42,221 @@ const showMessage = (message, sender) => {
...
 
-// Starts here
+// The dialogue starts here
 const greeting = () => {
-  showMessage(`Hello there, What's your name?`, 'bot')
-  // Just to check it out, change 'bot' to 'user' here 👆
+  showMessage(`Hi there! What's your name?`, 'bot')
+}
+
+const moodQuestion = () => {
```
IdaAspen, line 54: Suggestion to remove the start button after being used so that you cant press it when the bot has started. This also applies for the text input when you haven't pressed the start button.

code/script.js
```diff
@@ -26,22 +42,221 @@ const showMessage = (message, sender) => {
...
+        <option value="" selected disabled>Select an option 👇</option>
+        <option value="yes">Yes</option>
+        <option value="no">No</option>
+      </select>
+    `
+    const select = document.getElementById('select')
+    select.addEventListener('change', handleNeutralActivity)
+  }  else if (mood === 'worried') {
+    showMessage('That must be tough. Would you like to do a guided meditation?', 'bot')
+    inputWrapper.innerHTML = `
```
IdaAspen, line 111: Maybe a next step would be to remove the innerHTML when user has made its choice. Because now you can press yes, and then no, and then yes and so on.

code/script.js
```diff
@@ -26,22 +42,221 @@ const showMessage = (message, sender) => {
...
+}
+
+// Event Listeners
+
+startButton.addEventListener('click', () => setTimeout(greeting, 1000))
+form.addEventListener('submit', handleGreeting)
+
+
+
+// Pseudocode
```
IdaAspen, line 230: Love the pseudo code, a suggestion is to put it in the start of the code. So the user understands the flow better.

code/script.js
```diff
@@ -26,22 +42,221 @@ const showMessage = (message, sender) => {
...
+
+// Handle user input answers
+
+const handleGreeting = (event) => {
+  event.preventDefault()
+  const userName = inputValue.value
+
+  // Check whether input box is empty, interrupt chat if that's the case 
+
+  if (userName){
```
IdaAspen, line 143: Great to see some exception handling when user doesn't type anything!

code/style.css
```diff
@@ -74,6 +93,11 @@ main {
   display: flex;
   margin: 16px 8px 0 0;
   flex-shrink: 0;
+  /* visibility: hidden; */
```
IdaAspen, line 96: Just curious, as we are not completely sure ourselves what is the best option to use: Was there any reason why you used visibility: hidden instead of display: none?
We have understood from Maks that display: none is higher up in the hierarki because it is a stronger way to remove content.

IdaAspen (approved):
Great job, it was a very good chat bot and you should be proud of yourselves!
In our opinion you fulfilled at least the blue and red requirements 🥳

What we thought about in general was that you used different spacing in between the lines in the code and we couldn't see that it was consistent. Maybe something to think about in teh future?

Once again - well done, it was a really hard project!
//Sabrina & Ida

code/script.js
isomoth: Oh right! It's easy to miss these things.

code/script.js
isomoth: Right! Important UX details.

code/script.js
isomoth: Indeed, good catch! We got the same feedback from Jennie.

code/script.js
isomoth: Totally agree. Good tip!

code/style.css
isomoth: As we understood [from Stack Overflow](https://stackoverflow.com/questions/133051/what-is-the-difference-between-visibilityhidden-and-displaynone), if you use visibility:hidden the space that the element was occupying will still be preserved. While when using display: none, other elements will fill in that "empty" space. 

We were testing visibility:hidden to hide the bot's initial message and activate it through the start button, but later realized it was easier to do through an event listener :)

isomoth: Thank you so much for this insightful review, lots of useful things to think about!

IdaAspen: It was a pleasure! 
Oh that totally makes sense with the "empty" space - thanks for the explanation and for the link :)

PR #56, opened 2021-09-12
Katie & Pinar  -  DJ bot

pdetli:
https://djbot-katie-pinar.netlify.app/

We developed a music bot which  asks your mode and offers a song depending on its own mode. 

We proceeded according to the given iterations. The first step was adding a start button to activate the greeting. After the users first input, we proceeded to the next questions in turn. It changes the background color depending on the users input and offers alternative music styles. At the end of the conversation, it plays a song. When we finished the structure then we styled it.

We wanted to keep it simple so we wouldn’t get confused at the beginning of our js learning.  
If we had more time, probably the bot could have a better logical conversation with more functions.


code/script.js
```diff
@@ -1,47 +1,176 @@
...
 
-// Global variables, if you need any, declared here
-
-// Functions declared here
+//FUNCTIONS
 
 // This function will add a chat bubble in the correct place based on who the sender is
 const showMessage = (message, sender) => {
   if (sender === 'user') {
+    console.log("Hello!")
```
Lundgreneddie, line 16: I think we're suppose to remove all console.logs before handing in? :D I (we) forgets as well heh

code/script.js
```diff
@@ -1,47 +1,176 @@
...
 }
 
-// Set up your eventlisteners here
+sendBtn.addEventListener('click', () => {
+  userReply(userInput.value)
+  setTimeout(pickColor, 1000)
+})
+
+// Question 2 
+const pickColor = () => {
```
Lundgreneddie, line 100: Cool effect, we really liked this addition :) nice touch!

Lundgreneddie (approved):
Well done! Nicely structured, easy to follow the code and to understand what the different part do. 
Best regards, 
Sofia & Eddie :)

code/script.js
```diff
@@ -1,47 +1,176 @@
...
+  } else {
+    alert('Not a valid option!')
+  }
+}
+
+//to update the option for music style Party
+const updateParty = () => {
+  const selectParty = document.getElementById('selectParty')
+  if (selectParty.value === 'ninety') {
+    userReply(`90’s party`)
```
Lundgreneddie, line 65: When using the bot, maybe these should be in order based on the century :)

pdetli: Yeah that's right. I didn't realize until you said it :)

code/script.js
pdetli: You're right! There are always bunch of things we forget to do or not to do :)

PR #66, opened 2021-09-13
The chatbot

AschwinSiaila:
https://inspiring-bose-225fb3.netlify.app/

It was a learning process that I will never forget. Different kind of learn methods and a lot of learning materials. I had difficulties with the definitions of the code but I see some logic in it. It is still much and I am happy with the chatbot.

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -44,4 +44,4 @@ const greeting = () => {
 // But if we want to add a little delay to it, we can wrap it in a setTimeout:
 // setTimeout(functionName, timeToWaitInMilliSeconds)
 // This means the greeting function will be called one second after the website is loaded.
-setTimeout(greeting, 1000)
+setTimeout(greeting, 1500)
--- a/code/script.js
+++ b/code/script.js
@@ -1,53 +1,235 @@
-// All the DOM selectors stored as short variables
 const chat = document.getElementById('chat')
+const form = document.getElementById('name-form')
+const inputValue = document.getElementById('name-input')
+const sbmButton = document.getElementById('send-btn')
+const inputWrapper = document.getElementById('input-wrapper')
 
-// Global variables, if you need any, declared here
+let currentQuestion = 0
 
-// Functions declared here
+const botReply = (msg) => {
+     showMessage(msg, 'bot')
+   }
+  
+const userReply = (msg) => {
+     showMessage(msg, 'user')
+   }
 
-// This function will add a chat bubble in the correct place based on who the sender is
-const showMessage = (message, sender) => {
-  if (sender === 'user') {
-    chat.innerHTML += `
-      <section class="user-msg">
-        <div class="bubble user-bubble">
-          <p>${message}</p>
-        </div>
-        <img src="assets/user.png" alt="User" />  
-      </section>
-    `
-  } else if (sender === 'bot') {
-    console.log('this is here')
-    chat.innerHTML += `
-      <section class="bot-msg">
-        <img src="assets/bot.png" alt="Bot" />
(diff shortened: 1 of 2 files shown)


code/script.js
```diff
@@ -1,47 +1,235 @@
...
+    `
+    document 
+      .getElementById('food')
+      .addEventListener('click', () => {askForType('food')})
+    document
+      .getElementById('greetings')
+      .addEventListener('click', () => {askForType('greetings')})
+    document
+      .getElementById('jobs')
+      .addEventListener('click', () => {askForType('jobs')})
```
vrill, line 78: Clean way to add clarity to the code by using new lines. I like it! 👍

code/script.js
```diff
@@ -1,47 +1,235 @@
...
+
+      }else if (subject === 'greetings') {
+          showMessage(`I choose ${subject}`, 'user')
+          showMessage(`Ok, greetings it is! translate: Hello, How are you in Swedish?`, 'bot')
+     
+          inputWrapper.innerHTML = `
+          
+          <button id="good">Hej hur mår du?</button>
+          <button id="false">Hallo hoe gaat het?</button>
+          <button id="false0">Var bor du?</button>
```
vrill, line 153: When there are more than two choices, but only two logical outcomes ("good" or "false"), we could look into using `class="good"` and `class="false"` instead of `id`. This would help reduce the number of lines, specifically 136-138; 177-179; 199-201, where a single `.getElementsByClassName` could be used for both wrong answers.

code/script.js
```diff
@@ -1,47 +1,235 @@
...
+
+      showMessage(`My name is ${name}`, 'user')
+      inputValue.value = ''   
+
+      setTimeout(() => languageQuestion(name), 1500)
+    }
+
+   
+    const askForType = (subject) => {
+      if (subject === 'food') {
```
vrill, line 101: Impressed that most of the bot<>user conversation happens within a single function. Really cool to see and seems to help keep everything in order, making it a bit easier to understand what will happen.
However, a few descriptions/comments in the code would help explain the function and its stages/flow, as it initially is a bit to grasp.

code/script.js
```diff
@@ -1,47 +1,235 @@
...
+        .getElementById('false')
+        .addEventListener('click', () => {showMessage('Sorry try again' , 'bot')})
+        document
+        .getElementById('false0')
+        .addEventListener('click', () => {showMessage('Sorry try again' , 'bot')})
+
+
+
+        currentQuestion++
+        console.log('Current question is now', currentQuestion)
```
vrill, line 143: Really good use of `console.log` to keep track of where you are during development. Perhaps we can comment these out for the final product.

code/script.js
```diff
@@ -1,47 +1,235 @@
...
+          <button id="end">End Programming</button>
+              `
+              
+
+              document
+              .getElementById('retry')
+              .addEventListener('click', () => {location.reload()})
+              document
+              .getElementById('end')
+              .addEventListener('click', () => {showMessage('Congratulations on completing the program!', 'bot')
```
vrill, line 218: When most of the questions, answers and final outcomes are within the single `askForType` function, it has the downside of creating a lot of repetition. To reduce that it would probably be worth looking into breaking the stages of answers/questions out into their own functions to be called upon only when needed.

vrill (commented):
This is a really good chatbot with a simple and clear way of communication. It just works and I like that. I was very impressed that most of the communication happened within a single function, reducing the reliance on multiple functions - but adding some repetition in the process.

Good job, really looking forward to see more of what you will create! 👍

PR #85, opened 2022-02-04
Arianna and Åsa - chatbox

ariallahyar:
Arianna and Åsa worked together on this week 4 project. 
Link: https://hello-friend-chatbot.netlify.app/


code/script.js
```diff
@@ -1,47 +1,160 @@
...
+}
+
+// End message
+const endMessage = (message) => {
+  showMessage(`Alright, let's keep in touch. Have a good day and don't forget that you're a rockstar ${message}!`, 'bot');
+  inputWrapper.innerHTML = '';
+}
+
+// Reusable event listener(s)
+const addButtonListeners = (buttonClassName) => {
```
jjwwkkmm, line 125: Perhaps remove the parameter buttonClassName and write ```const sendButtons = document.getElementsByClassName(.'button-input');``` diretcly on line 126 and then invoke 
the function with ```addButtonListeners()``` only on line 59 to increase readability

code/script.js
```diff
@@ -1,47 +1,160 @@
...
+  const sendButtons = document.getElementsByClassName(buttonClassName);
+  console.log(sendButtons)
+  
+  Array.from(sendButtons).forEach(button => {
```
jjwwkkmm, line 129: Is there a reason for using Array.from()? :) Think ```sendButtons.forEach```would work on line 129, but then you might need to change line 126 above to ```const sendButtons = document.querySelectorAll(.'button-input);``` first

code/script.js
```diff
@@ -1,47 +1,160 @@
...
+    setTimeout(endMessage, 1000, nameInput.value);
+  }
+}
+
+// Question 1
+const question1 = () => {
+  showMessage(`Hello there, what's your name?`, 'bot')
+
+  document.getElementById('name-form').onsubmit = event => {
+    event.preventDefault();
```
savannah-hayes, line 43: Do you think it is better to move the onsubmit listener outside of the question1 function? I think it could be moved to the bottom of your JS file where you have the other event listeners.

code/script.js
```diff
@@ -1,47 +1,160 @@
...
+    form.submitButton.disabled = true;
+    showMessage(nameInput.value, 'user');
+    questionGenerator();
+  }
```
savannah-hayes, line 47: Do you think it is better to move the onsubmit listener outside of the question1 function? Do you think it could be moved to the bottom of your JS file where you have the other event listeners?

code/script.js
```diff
@@ -1,47 +1,160 @@
...
-// This function will add a chat bubble in the correct place based on who the sender is
+// Function to move through questions
+const questionGenerator = () => {
+  questionCounter++;
+  if (questionCounter === 1) {
+    setTimeout(question1, 1000);
+  } else if (questionCounter === 2) {
+    setTimeout(question2, 1000, nameInput.value);
+  } else if (questionCounter === 3) {
+    setTimeout(question3, 1000, buttonText);
```
jjwwkkmm, line 28: So buttonText is the same parameter that's passed as parameter in line 63 but called answer there instead, in which buttonText is a variable described on line 132 but declared on line 10 as a global variable. What if we move line 132 out as a global variable, change line 131 to ```showMessage('buttonText', 'user')```, remove buttonText parameter in line 28, remove answer parameter in line 63, and changes line 65 to ``` if (buttonText ==='Great!')```. Would that work?

code/script.js
```diff
@@ -1,47 +1,160 @@
...
+  const textInput = document.getElementById('text-input');
+  document.getElementById('send-btn').addEventListener('click', () => {
+    showMessage(textInput.value, 'user');
+    textInput.value ='';
+    questionGenerator();
+  }, {once: true});
+}
+
+// Question 4
+const question4 = (answer) => {
```
jjwwkkmm, line 84: Seems like the answer parameter is never used here, buttonText is already defined as a global variable :)

code/script.js
```diff
@@ -1,47 +1,160 @@
...
+    <input id="text-input" type="text" />
+    <button id="send-btn">Send</button>
+  `
+  const textInput = document.getElementById('text-input');
+  document.getElementById('send-btn').addEventListener('click', () => {
+    showMessage(textInput.value, 'user');
+    textInput.value ='';
+    questionGenerator();
+  }, {once: true});
+}
```
savannah-hayes, line 97: It seems that the code on lines 75-81 and 91-97 are duplicated. I would suggest maybe moving the duplicated code into a function you can call into both question3 and question4.

code/script.js
savannah-hayes: Yes. I am thinking the same thing with Array.from( ) ? Were you trying to accomplish something like this? For example something similar to what you did on lines 75-80 and 91-96.

```
const question2 = (message) => {
  showMessage(`How is your day going, ${message}?`, 'bot');

  inputWrapper.innerHTML = `
    <button id="button-input" class="button-input">Great!</button>
    <button id="button-input" class="button-input">Could be better</button>
    `
  const buttonInput = document.querySelectorAll('#button-input');

  buttonInput.forEach(button => {
    button.addEventListener('click', () => {
      showMessage(button.innerText, 'user');
      buttonText = button.innerText;
      questionGenerator();
    }, { once: true });
  })
}
```

savannah-hayes: It seems this reusable event listener you have in your comment on line 124. Is it used more that once? Am I mistaken isn't it only used inside the question2 function?

code/script.js
```diff
@@ -1,47 +1,160 @@
...
+  startContent .style.display = "none";
+  chatBox.style.display = "flex";
+  questionGenerator();
+});
 
-// This function will add a chat bubble in the correct place based on who the sender is
+// Function to move through questions
+const questionGenerator = () => {
+  questionCounter++;
+  if (questionCounter === 1) {
```
savannah-hayes, line 23: Great job adding a counter increment! And clever way to using it here in your questionGenerator function.

code/script.js
```diff
@@ -1,47 +1,160 @@
...
+
+// Question 1
+const question1 = () => {
+  showMessage(`Hello there, what's your name?`, 'bot')
+
+  document.getElementById('name-form').onsubmit = event => {
+    event.preventDefault();
+    form.submitButton.disabled = true;
+    showMessage(nameInput.value, 'user');
+    questionGenerator();
```
savannah-hayes, line 46: I liked how you both thought to disable the button after the form submission. :)

code/script.js
```diff
@@ -1,47 +1,160 @@
...
+  showMessage(`Sounds great! When should we chat next?`, 'bot');
+  inputWrapper.innerHTML = `
+    <select id="option-menu" type="option">
+      <option selected disabled>Choose a time</option>
+      <option>Tomorrow</option>
+      <option>In three days</option>
+      <option>Next week</option>
+      <option>I'm not sure</option>
+    </select>
+    `
```
savannah-hayes, line 110: Great job reusing the inputWrapper div instead of creating new div wrappers for each of the buttons, inputs, and select. Way to keep your code dry!

code/index.html
```diff
@@ -12,20 +12,26 @@
...
 
   <body>
-    <h1>Welcome to xxx.</h1>
-    <h2>Ask about xxxx.</h2>
-    <main>
-      <section class="chat" id="chat"></section>
+    <div id="start-content" class="start-content">
+      <h1>Hello friend</h1>
+      <button id="start-button" class="start-button">Wanna talk?</button>
+    </div>
```
savannah-hayes, line 18: Beautiful start page!

savannah-hayes (approved):
Your code follows all the requirements and it seems that you both have been able to even surpass the general requirements. I really liked how you included a start page, it was beautifully done! The code works as it should and nothing is broken. 

The code seems to have some duplicates that could be tidied up. But overall great job with your chatbot and clever idea for a chatbot. Great use of the all the inputs, buttons, and date select element.

jjwwkkmm (commented):
I also want to add that being the first week of Javacscript this is impressive, as always there are things that could be improved for everyone but that do not indicate the current work is not good. I believe if you had more time to finish the project, the next steps would be to see over readability and make the code even more clean. As Savannah also said, the site looks beautiful and the chatbot is working, so next steps would be to refactor the code.

Well done!

PR #91, opened 2022-02-05
Project chatbot, Tiina Liukkonen & Emma Högberg

tiiliu:


LovisaBrorson:
Hi!

Smart way to line up the question in an "if" on row 46.
The code is easy to follow and has a good pattern. It's simple and easy to read.

Great css! And fun with the sound effect. 

Well done!
Anki & Lovisa

PR #101, opened 2022-02-07
Chatbot Dorothea

Dorothea-B:
https://silly-hoover-8e96d6.netlify.app/


code/script.js
```diff
@@ -30,14 +45,181 @@ const showMessage = (message, sender) => {
...
+
+
 // Starts here
+
+//QUESTION 1
 const greeting = () => {
-  showMessage(`Hello there, What's your name?`, 'bot')
+  questionNumber = 1
+  showMessage(`Hi, let's start with your name?`, 'bot')
   // Just to check it out, change 'bot' to 'user' here 👆
```
MarianneArdin, line 83: We can't type our names in the input field.

code/script.js
```diff
@@ -30,14 +45,181 @@ const showMessage = (message, sender) => {
...
+  nextQuestion(input.value)
+ })
+
+
+//QUESTION 2
+
+const showTattooStyles = (msg) => {
+  questionNumber = 2
+  botReply(
+    `Welcome, ${msg}. What tattoo style are you looking to get?`
```
MarianneArdin, line 94: We get this question twice.

code/script.js
```diff
@@ -30,14 +45,181 @@ const showMessage = (message, sender) => {
...
+
+//QUESTION 2
+const thankYou = () => {
+  botReply(`Thank you for your booking!`)
+  inputWrapper.innerHTML = ``
+}
+
+
+
 // Set up your eventlisteners here
```
MarianneArdin, line 234: We think the reason you are not able to enter name in first box is because the event listeners (line 215-221) are outside of the functions so the will activate, when we commented out these lines, the bot worked.

MarianneArdin (changes requested):
Good work Dorothea, we understand this week was a struggle. The bot looks great. We have commented in the code for the event listeners that prevent the code from working. If you comment out OR delete line 215-221 in JS the bot will work. Try this change and see if the bot works as it should to meet minimum requirements. //Reviewed by Marianne and Kristiina.

PR #108, opened 2022-08-26
Kristin Larsson & Caroline Söderström

SoderstromC:
link: https://animated-lebkuchen-1f1c42.netlify.app/


code/script.js
```diff
@@ -33,19 +46,124 @@ const showMessage = (message, sender) => {
...
+
+  if (choice === '1') {
+    showMessage("Oh! We see you brought your best company, Yourself! We'll make sure you get the best table!", 'bot')
+  } else if (choice === '5') {
+    showMessage("We love a big crowd! Please contact us at gm@fakemail.com for large bookings!", 'bot')
+  } else {
+    botAnswer("We look forward to your visit!", 'bot')
+  }
+  
+  select.addEventListener('change', () => nextQuestion(inputValue))
```
petrasson, line 166: we get an error message on this select.addEventListener. We think you are trying to use the same constant used on line 148. Perhaps you can use let instead or create a new variable for this AddEventListener.

agadman (commented):
Great job! The chatbot is very nice and personal, good job with the CSS - very nice colors! The readme doc was very short and concise. Except for the error on row 166 (see comment), the javascript code works well and the flow is easy to follow.  We have added a few comments and thoughts in the code.

code/script.js
```diff
@@ -33,19 +46,124 @@ const showMessage = (message, sender) => {
...
+      <option value='' selected disabled> Choose time </option>
+      <option value='Lunch'>Lunch</option>
+      <option value='After Work'>After Work</option>
+    </select> 
+    `
+  } 
+  else {
+    inputWrapper.innerHTML=`
+    <select id = 'select'>
+      <option value='' selected disabled> Choose time </option>
```
agadman, line 119: Try to remove selected before disabled. We think it should only be disabled on this option. Code seems to work, but it might cause an error.

PR #111, opened 2022-08-26
Chatbot - Tina Bruce & Linnéa Ajger

LinneaAjger:


code/script.js
```diff
@@ -1,11 +1,118 @@
...
+    </select>
+    `
+    question++; 
+    document.getElementById('select').addEventListener('change', () => setTimeout(() => showMessage(select.value, 'user'), nextQuestion()));
+  }
+
+  const showBaldOptions = () => {
+    showMessage("I woke up bald!", 'user');
+    const wig = `<a href=https://www.wig.se/>wig.se</a>`
+    setTimeout(showMessage(`Sorry we cant help you 🤷‍♀️. Perhaps you can get help at ${wig}`, 'bot'), 2000);
```
fannycardell, line 67: For all buttons there is a delay of 2000 milliseconds, which is very nice, but it seams that it has fallen out in this section because the answer from the bot appears at the same time as user.

code/script.js
```diff
@@ -1,11 +1,118 @@
...
+
+    question++; 
+    document.getElementById("long")
+    .addEventListener('click', () => setTimeout(() => showMessage('Long hair it is!', 'user'), nextQuestion(), 1000)); 
+    document.getElementById("short")
+    .addEventListener('click', () => setTimeout(() => showMessage('Short hair it is!', 'user'), nextQuestion(), 1000));    
+
+  }
+
+  const timeSlots = () => {
```
fannycardell, line 86: Incredibly nice done with this drop list of time selections!

fannycardell (commented):
Requirements for the project are fulfilled and the site works as it should and the code itself is easy to follow and well written, super nice job!
Nicely named functions, easy to follow. Would be helpful with some "outcomment" headlines in the code for the reader to follow were in the conversation the bot and user are. Super fun with the new image of the bot! Would be fun to have the title with HairSOS=)

PR #113, opened 2022-08-26
Emil & Jessika

EmilBergvik:
https://moviesuggestor.netlify.app/

Made by: Emil Bergvik Jessika Lind


code/script.js
```diff
@@ -1,51 +1,285 @@
...
-
-// If you need any global variables that you can use across different functions, declare them here:
-
-
-// Declare your functions after this comment
+// All the DOM selectors stored as short variables
+const chat = document.getElementById('chat')
+const form = document.getElementById('name-form')
+const inputWrapper = document.getElementById('input-wrapper')
+let userName = ""
```
brucette, line 5: This variable doesn't seemed to be used anywhere in the code/is not needed as a global variable, as it is declared on row 38, handleResponse function.

code/script.js
```diff
@@ -1,51 +1,285 @@
...
+        <img src="assets/icon-bot.png" alt="Bot" />
         <div class="bubble bot-bubble">
           <p>${message}</p>
         </div>
       </section>
     `
   }
   // This little thing makes the chat scroll to the last message when there are too many to be shown in the chat box
-  chat.scrollTop = chat.scrollHeight;
+  setTimeout (() => chat.scrollTop = chat.scrollHeight, 1)
```
brucette, line 29: We are just curious, what is the function of this line? Perhaps a comment explaining it would be good?

code/script.js
```diff
@@ -1,51 +1,285 @@
...
 }
 
-// Starts here
+//Question 1: Bot greets and asks for the user's name
```
brucette, line 32: Very nice way of explaining your code, make's it easier to read and follow!

code/script.js
```diff
@@ -1,51 +1,285 @@
...
+      inputWrapper.innerHTML = ''
+      setTimeout (() => chooseByMood('optAngry'), 1000)
+    })
+  }
+  } 
+  
+  else if (selection === 'option2') {
+    showMessage(`Well, now that you know it works, you're more than welcome back once you need my help!`, 'bot')
+    setTimeout (() => inputWrapper.innerHTML=`
+    <p>Thanks for using this bot!</p>
```
brucette, line 105: Doing a separate function for the Thank you message might be a good idea as it is repeated in several of you functions.

code/script.js
```diff
@@ -1,51 +1,285 @@
...
+      showMessage(`I'm feeling sad today`, 'user')
+      inputWrapper.innerHTML = ''
+      setTimeout (() => chooseByMood('optSad'), 1000)
+    })
+
+    document.getElementById("optAngry").addEventListener('click', () => {
+      showMessage(`I'm feeling angry today`, 'user')
+      inputWrapper.innerHTML = ''
+      setTimeout (() => chooseByMood('optAngry'), 1000)
+    })
```
brucette, line 98: Perhaps having the moodChoice function defined outside the doYouNeedHelp() function would make the if-else statements easier to follow?
Or maybe it's a matter of taste?

code/script.js
```diff
@@ -1,51 +1,285 @@
...
+        inputWrapper.innerHTML = ''
+        setTimeout (() => lastReplyHappy('singingInTheRain'), 1000)
+      })
+
+      document.getElementById("greaseMovie").addEventListener('click', () => {
+        showMessage(`I'll watch Grease tonight`, 'user')
+        inputWrapper.innerHTML = ''
+        setTimeout (() => lastReplyHappy('grease'), 1000)
+      }) 
+    }
```
brucette, line 140: As with the earlier comment of row 75, we would also consider putting this in a separate function to ease reading.

code/script.js
```diff
@@ -1,51 +1,285 @@
...
+        showMessage(`I'll watch The Notebook tonight!`, 'user')
+        inputWrapper.innerHTML = ''
+        setTimeout (() => lastReplySad('theNotebook'), 1000)
+        })
+
+        document.getElementById("brokebackMountain").addEventListener('click', () => {
+        showMessage(`I'll watch Brokeback Mountain tonight`, 'user')
+        inputWrapper.innerHTML = ''
+        setTimeout (() => lastReplySad('brokebackMountain'), 1000)
+        })
```
brucette, line 170: Same comment as row 75. However, you are very structured and consistent with your code!

code/script.js
```diff
@@ -1,51 +1,285 @@
...
+          showMessage(`I'll watch Terminator 2 tonight!`, 'user')
+          inputWrapper.innerHTML = ''
+          setTimeout (() => lastReplyAngry('terminator2'), 1000)
+        })
+  
+        document.getElementById("djangoUnchained").addEventListener('click', () => {
+          showMessage(`I'll watch Django Unchained tonight`, 'user')
+          inputWrapper.innerHTML = ''
+          setTimeout (() => lastReplyAngry('djangoUnchained'), 1000)
+        })
```
brucette, line 201: Same comment as above :)

code/script.js
```diff
@@ -1,51 +1,285 @@
...
+        document.getElementById("djangoUnchained").addEventListener('click', () => {
+          showMessage(`I'll watch Django Unchained tonight`, 'user')
+          inputWrapper.innerHTML = ''
+          setTimeout (() => lastReplyAngry('djangoUnchained'), 1000)
+        })  
+    }
+  }
+}
+
+//The last reply from bot: The bot replies to the user's choice of films, and thanks the user for using the chat bot.
```
brucette, line 206: Great examples of easy to follow if-else statements on rows 207 - 271!

code/script.js
```diff
@@ -1,51 +1,285 @@
...
+    setTimeout (() => inputWrapper.innerHTML = `
+    <p>Thank you for using this bot!</p>
+    `, 1000)
+  }
 }
 
 // Set up your eventlisteners here
+form.addEventListener('submit', (event) => {
+  event.preventDefault();
+  const value = document.getElementById('name-input').value;
```
brucette, line 277: Just wondering ourselves if we should avoid naming variables with the same name as properties that exist in javascript..(i.e. const value vs. .value) ?

brucette (commented):
Good, professional looking bot! Flows really nicely, with no extra delays in the responses. 
We like that the bot is very personalised with different answers for every choice - an interesting and funny bot!

Code is well described with comments and the naming of variables and functions, making it easy to understand. Good consistent structure. However, we suggest it could be made easier to read if some of the code was to be made into their own functions.

Edit by Tina Bruce - just to add that review was done together with Linnéa Ajger :)

code/script.js
jessikalind: Good point!

code/script.js
EmilBergvik: That would probably look much better code-wise, thanks! 😊

code/script.js
jessikalind: We added the setTimeout to 1 on the chat.scrollTop to make sure it always scrolls down to the bottom when a new element is loaded. For more information about this, see this question on Stack Overflow: https://stackoverflow.com/c/technigo/questions/3932. 
I agree that we could have added a comment about this 😊

code/script.js
jessikalind: Very good point!

code/script.js
jessikalind: Sure!

code/script.js
jessikalind: I agree that it would be less confusing if we would have chosen another name for it! 

However, it seems that this is widely used (for example in Technigo's own example in Sprint 1). So, I don't know, maybe we just have to get used to it 🤔

PR #115, opened 2022-08-26
Team hippo, Linda och Neah, project-chatbot

Archkrull:


katirum (commented):
-  Am I able to understand the code easily?
Yes, we understand the code and they wrote comments for us to read easier.

-  Does the code have lots of duplicates? Could it be tidied up?
We think that their code is written very neatly.

-  Does it work or does anything look broken?
No, it definitely works as intended.

-  Does it follow the general and blue requirements?
Yes, they followed the requirements.

Our comment regarding the chatbot:
We think it is really creative way of creating the chatbot. We learnt something new from this bot, for example how they put in credit for the image owner. We like this chatbot a lot.

//Kartika & Lisa

code/script.js
```diff
@@ -34,18 +39,139 @@ const showMessage = (message, sender) => {
 }
 
 // Starts here
+//Question number 1
 const greeting = () => {
   // here we call the function showMessage, that we declared earlier with the argument "Hello there, What's your name?" for message, and the argument "bot" for sender
```
katirum, line 44: It's nice how they re-describe the comment instead of keeping the original one.

code/script.js
```diff
@@ -34,18 +39,139 @@ const showMessage = (message, sender) => {
...
+  showMessage(tourGuide, 'user');
+  setTimeout(() => chooseActivity(tourGuide), 750);
+}
+
+// Question numer 3, drop down menu for choices
+const chooseActivity = (tourGuide) => {  
+  showMessage (`You have chosen the ${tourGuide}. What do you want to do together?`, 'bot');
+  if (tourGuide === 'Pirate') {
+    document.body.style.backgroundImage = "url('assets/pirate-ship.jpg')"
+    // "Pirate Ship. Nikon D3100. DSC_0170." by Robert.Pittman is licensed under CC BY-ND 2.0.
```
katirum, line 90: This part is really cool and creative. Putting new background for each options. Also, from this we can learn that it's possible to do this way.

PR #117, opened 2022-08-26
Project chatbot

uohman:


uohman:
Project by Ulrika Öhman & Kaja Wilbik

RosenholmJohanna:
The chatbot works well and it has a clear layout kind to the users’ eyes. Could be nice with some sound effect, but we really like the cursor and the chat-icons.  
Names of functions are good and it’s easy to follow the code and understand it. Your own comments whitin the code is good and as a reviewer it makes it easy to understand what different parts of the code is doing what. There’s is only a few of your comments that might can be removed, just to make it even more clean (see comments).  But in general, we can’t see anything that is missing or wrong and the links works fine. Nice job! 
// Johanna Rosenholm & Maria So

PR #123, opened 2022-08-28
Kim & Jacob

Earnieball:
Week 1: Chatbot


AgnesAhlman (commented):
* Delete the code that is not relevant to make it easier for the eye. 
* We were able to understand the code and the structure easily. 
* Nice that the input requires a text to move on. 
* It’s just a minor detail that could improve readability. You could separate the `document.getElementById('Pale Ale-btn')` from the adding of the event listener by defining your buttons as variables first: e.g `const paleAleBtn = document.getElementById('Pale Ale-btn')`
* Avoid using spaces when defining Ids such as `Pale Ale-btn`.

PR #129, opened 2022-08-28
W1-ProjectChatbot-SofieRydmark-CarinaMoreno-JessicaMo

SofieRydmark:
Link to project: https://hotelvosje-chatbot-lions.netlify.app/ 

Project done by:
Sofie Rydmark
Jessica Mo
Carina Moreno 
(from team Lions)


code/script.js
```diff
@@ -1,51 +1,312 @@
...
+    showMessage(`Please, give me your room number.`, 'bot');
+  } else {
+    greetingLoop(room);
+  }
+}
+
+// The bot answers hello room #. This const will also be used as loop back selections if guest needs something else
+
+const greetingLoop = (room) => {
+  typing();
```
KajaWilbik, line 103: What means this "typing"? Is it to make look like a bot it's thinking and typing?

code/script.js
```diff
@@ -1,51 +1,312 @@
...
+  showMessage('What can I help you with?', 'bot');
+  inputWrapper.innerHTML = `
+  <button id="amenities" type="sbumit" class="chat-btn">Amenities </button>
+  <button id="room-service" type="submit" class="chat-btn">Room service </button>
+  <button id="wake-up" type="submit" class="chat-btn">Wake-up call </button>
+  `
+  document.getElementById('amenities').addEventListener('click',() => {
+    showMessage(`I need amenities, please`, 'user')
+    setTimeout( ()=> {
+      preloader.style.display = "none";
```
KajaWilbik, line 121: What is style.display doing? Why not only setTimeout? Maybe some comment here?

code/script.js
```diff
@@ -1,51 +1,312 @@
...
+
+}
+
+// Question 3 - Bot asks what guest needs depending on what guest chose in helpselection
+
+const selectionAnswer = (selection) => {
+
+  if (selection === 'amenities') {         // If guest needs amenities 
+    showMessage(`Sure, what do you need?`, 'bot');
+    inputWrapper.innerHTML = `
```
KajaWilbik, line 144: The page scrolling does not show a message properly after choosing amenities. The buttons are bigger and overlap the main window.

code/script.js
```diff
@@ -1,51 +1,312 @@
...
+  document.getElementById('no').addEventListener('click', () => {    // If guest says no 
+    showMessage(`No, thank you`, 'user')
+    setTimeout(() => confirmation(), 1500)
+
+  })
+
+    
+  }
+
+  else if (selection === 'wake-up') {
```
uohman, line 267: Why not use "else" instead of "else if"?

code/script.js
```diff
@@ -1,51 +1,312 @@
...
-  // Just to check it out, change 'bot' to 'user' here 👆
+      showMessage("Welcome to Hotel Vosje. What is your room number?", 'bot'),2000; 
+    };
+  
+
+// User answers with room number
+
+const handleFormInput = (event) => {        // at submit this function will be invoked
+  event.preventDefault();        // prevents website refresh at submit
+  const room = roomInput.value;        // input value will be stored in the const name
```
uohman, line 89: Could it be a good idea to block that the bot takes strings, so that it only takes numbers? Right now you can write your name instead of room number and it still works.

KajaWilbik (approved):
Great job! Your code is really easy to follow and neatly done. Our comments are only suggestions.

uohman (approved):
Great job! See Kaja's comments. :)

code/script.js
SofieRydmark: yes the typing() is to invoke the function that was created before that to make it look like the bot is thinking and typing here :)

code/script.js
SofieRydmark: oh i think this is actually an error and this "preloader.style.display ="none"" should not be there. it was from when we tried to make the bot-typing function works and it seems we forgot to remove it there :)

code/script.js
SofieRydmark: ok thanks we will look at that and try to fix it!

code/script.js
SofieRydmark: i wrote this part and i think it was (if i remember correctly) because if I only put "else" at first, it didnt work somehow. But it would probably be more best practice to just use "else"?

SofieRydmark: we will try to change to else and see if it works the same and probably use that then :)

code/script.js
SofieRydmark: good idea and good spotting. we actually talked about that but decided to leave it like is because maybe hotels have names or letters in the rooms combined. like "top suite 3" or "panther room" (lol) or something. :)

code/script.js
jessnmo: using else if is because we're giving it a condition, else returns a function that applies to everything that doesn't have the condition assigned to if

PR #134, opened 2022-08-29
Tara, Petra

Tara78:


code/newapp.js
```diff
@@ -0,0 +1,95 @@
+document.addEventListener('DOMContentLoaded', () => {
+  const chat = document.getElementById('chat');
+  const inputWrapper = document.getElementById('input-wrapper');
+  const nameInput = document.getElementById('name-input');
+  const sendButton = document.querySelector('.send-btn');
```
jnbjensen, line 5: Is there a reason why you chose querySelector here instead of getElementbyId? It works fine, we were just wondering. :)

code/newapp.js
```diff
@@ -0,0 +1,95 @@
...
+
+  // Greeting start
+  const greeting = () => {
+    // here we call the function showMessage, that we declared earlier with the argument "Hello there, What's your name?" for message, and the argument "bot" for sender
+    showMessage("Hello there, What's your name?", 'bot');
+
+  }
+
+  // Eventlisteners 
+  sendButton.addEventListener('click', (e) => {
```
jnbjensen, line 43: Might have been clearer to spell out 'event' here instead of just e?

jnbjensen (commented):
Hi Tara and Petra! Great job getting the code to work in the end! : ) As discussed with Tara, we reviewed the code in the newapp.js file, so our comments below are aimed at that one.

We tested the code and it works & meets all the requirements & guidelines for the project. Some general feedback:

1) We would recommend adding more comments to your code, both to increase readability and to make it easier for yourselves. For example, you could add a line above each function describing what exactly it does. 
2) Really nice that you added the document.addEventListener('DOMContentLoaded' function on line 1!
3) Well-named functions that do what they say. 
4) Well-structured code.

PR #140, opened 2023-02-10
Chatbot - Zoltar

khleoth:
Chatbot project by Leo Thunell


code/script.js
```diff
@@ -1,51 +1,199 @@
...
+  }
 }
 
 // Starts here
-const greetUser = () => {
-  // here we call the function showMessage, that we declared earlier with the argument "Hello there, What's your name?" for message, and the argument "bot" for sender
-  showMessage("Hello there, What's your name?", 'bot');
-  // Just to check it out, change 'bot' to 'user' here 👆
+const greeting = () => {
+  questionNumber = 1
```
Joanna-Philips, line 86: Since questionNumber is already declared above, it is probably not needed here.

code/script.js
```diff
@@ -1,51 +1,199 @@
...
+  } else {
+    botReply('Very well, be gone then.'),
+    setTimeout(location.reload(), 3000)
+  }
+
+  const select = document.getElementById('select')
+  select.addEventListener('change', () => nextQuestion('Tell me of my ' + select.value))
+}
+
+const chosenFortune = (type) => {
```
Joanna-Philips, line 134: The function chosenFortune does not seem to be working.

Joanna-Philips (changes requested):
Leo, well done on the project!
I think it follows  and exceeds the requirements  for this week. 

What I liked:
- the code was clean and easy to read, with variables and functions where appropriate, thus eliminating duplicates. 
- good use of different form elements (buttons, dropdown, text input)
- nice job on the hover effects for the buttons and the effects for the main headings. and the sound.

What could be better:
- The chosenFortune function does not seem to work.  It is always the last botReply ("Really, this again...") that is displayed no matter the selection.

code/script.js
khleoth: Hmmm, having a hard time making it work, I'll push the latest changes. I had a capital F on one and just f on another "future" but even though i changed that I cant seem to get it to work... any ideas what could be the problem?

Joanna-Philips: I think it might be the comparing string in (type === ' '). 
Try `(type === "Tell me of my future") ` `(type === "Tell me of my family")`  because that's the parameter we're passing to nextQuestion().

PR #142, opened 2023-02-10
HappyCloude/Project-Chatbot

EvaSjoholm:
https://tubular-conkies-f117c9.netlify.app/


EvaSjoholm:
Very challenging to do all solo as a beginner to IT in general and also being a creative soul.
I am okey with the outcome and proud that I finished it somewhat functional.

PR #145, opened 2023-02-11
Amandas chatbot

Amanda-Sjstrm:
My first JavaScript-project!


code/script.js
```diff
@@ -1,51 +1,153 @@
...
+  }
+
+  //You will love your new house
+  inputWrapper.innerHTML = `
+<button id="yayButton">Yay!</button>
+`;
+  document
+    .getElementById("yayButton")
+    .addEventListener("click", () =>
+      handleInput("I can't wait to become the gratest witch in the world!")
```
smirrebinx, line 141: This doesn't affect the code but there might be a typo in the message on line 141:
 Gratest = Greatest

code/script.js
```diff
@@ -1,51 +1,153 @@
...
-// But if we want to add a little delay to it, we can wrap it in a setTimeout:
-// setTimeout(functionName, timeToWaitInMilliSeconds)
-// This means the greeting function will be called one second after the website is loaded.
-setTimeout(greetUser, 1000);
+//4 - Bot will place you in Hogwarts house
+const houseStatement = (colour) => {
+  questionStep++;
+  if (colour === "Green") {
+    botAnswer(`SLYTHERIN!`);
+  } else {
```
smirrebinx, line 129: The if and else places the user in the same house: Slytherin, is that on purpose? 
Or should there be another housing option? Line 129

smirrebinx (commented):
Overall, very cool app. Good variable and function names make the code easy to follow and understand. Good use of comments explaining the functions. One suggestion was made regarding a possible typo and a question about an if else statement.

code/script.js
Amanda-Sjstrm: I wanted the user to be placed in Slytherin either way, but if you answer anything other than "green" the bot just has to think about it for a second first 😋

code/script.js
Amanda-Sjstrm: Yes its a typo 😅

PR #146, opened 2023-02-11
Nora Terbocs - Week 1 - chat bot

noraterbocs:


Yu-Miao222:
I wish I could give more suggestion, but your code is very neat and clear, nice and advanced. You have reached all level of goals required. Well done!!

PR #147, opened 2023-02-11
ChatBot

IrupePozoGraviz:
Want able to get it exactly the way I wanted becuse time and a fever


IrupePozoGraviz:
.

IrupePozoGraviz:
changed text

code/script.js
```diff
@@ -1,51 +1,270 @@
...
+
+// Removes bot typing effect
+
+const removeTyping = () => {
+  document.getElementById('preloader').style.display='none'
+}
 // This function will add a chat bubble in the correct place based on who the sender is
-const showMessage = (message, sender) => {
+//without the ${message} the actual message(parameter) in const greetUser will not be passed, only string message. Only the parameter 'bot' 
+function showMessage(message, sender) {
```
JohLeo, line 26: Tudelo!  Why did you change it from a const into a function? Qurious!

code/script.js
```diff
@@ -1,51 +1,270 @@
...
+
+};
 
-// When website loaded, chatbot asks first question.
-// normally we would invoke a function like this:
-// greeting()
-// But if we want to add a little delay to it, we can wrap it in a setTimeout:
-// setTimeout(functionName, timeToWaitInMilliSeconds)
-// This means the greeting function will be called one second after the website is loaded.
+//We welcome the user (based on current time) with a delay of 1s
```
JohLeo, line 63: You do not greet the user based on curent time here right, or am I missing something? The delay is 1 sek but I cant see a specific time-greeting. If so I wanna know how (cause I want to learn! 😬)

code/script.js
```diff
@@ -1,51 +1,270 @@
...
+
+
+
+wrapper.innerHTML =`
+<button id="fun-fact" > Fun Fact </button>)
+<button id="history" > History </button>
+<button id="order" > Send me a pineapple!! </button>`
+
+//why has this eventlistneer started to show up as a bubble?
+document.getElementById('fun-fact').addEventListener('click', () => userReplyOne('fun'))
```
JohLeo, line 94: Did you manage to fix this?

code/script.js
```diff
@@ -1,51 +1,270 @@
...
+document.getElementById('order').addEventListener('click', () => answerThree("order"))
+
+}
+//answer and first interaction
+
+const userReplyOne = (chooseOne) => {  
+if (chooseOne === 'fun'){
+
+
+  setTimeout (
```
JohLeo: i'm not sure here, but could the effect of this be better if you set the "setTimeout" to different times, so start with 1000, then 1500 and so on?  Since it should be reading the code chronological I think the setTimeOut should be different.

code/script.js
```diff
@@ -1,51 +1,270 @@
...
+
+answerTwo = (type) => {
+  console.log ('working1')
+  if (type === 'historyEnd'){
+
+    setTimeout ( () =>
+    botReply(`More history for the people: The pineapple spread throughout South America, the Caribbean, Central America and Mexico, where it was cultivated by the Mayas and the Aztecs. Columbus came across the pineapple in 1493 and took it back to Europe.`), 1500)
+
+    setTimeout ( () =>
+    botReply ('Are you ready to order your pineapple?'), 1000)
```
JohLeo: Same here with the time, the first botReply is set to 1500 and the second to 1000. This means the second "Are you ready to order your pinapple?" will show up before the history answe!

code/script.js
```diff
@@ -1,51 +1,270 @@
...
+
+
+
+}
+
+    
+    // This wont work?
+  const endForm = (pineapple) => {
+    setTimeout ( () =>
+    botReply (`Yey thanx for orderinge a ${pineapple } write your name and e mail and we will send you information on where to get your happy pineapple`), 1000)
```
JohLeo: Is it not working? I got it to work on most of the endings!

code/script.js
```diff
@@ -1,51 +1,270 @@
...
+document.getElementById('email').addEventListener('click', () => thankYou())
+document.getElementById('button').addEventListener('click', () => thankYou())
+} 
+
+//these wont work!
+const thankYou =() => {
+  setTimeout ( () =>
+  botReply ("Pineapples are international symbols of welcome and are a symbolic way of saying “you are perfect” when presented to someone."), 1000)
+setTimeout ( () =>
+botReply ('Thank you for ordering hope to see you again!'), 1000)
```
JohLeo: My guess for it not working is because your are not invoking it after the order is sent. You can fix this! Judging from all the other code above you know how to solve this!

code/style.css
```diff
@@ -147,3 +148,17 @@ button:hover {
...
+  border-radius: 4px 0 0 4px;
+  background: #e5e9ff;
+  color: #0026ff;
+  padding: 16px;
+  font-size: 16px;
+  font-family: 'Montserrat';
+  font-weight: 600;
+  line-height: 26px;
+  flex: 1;
+}
```
JohLeo, line 164: For your portfolios sake, should you maybe put some time into making it responsive with an media query for the messages (up the % on the bubbles) and maybe change the sizes of the buttons and input and set the main to another height and width (maybe less border-radius?)... this is a fast fix, you're a css code pro! 💪🏼

JohLeo (changes requested):
Woho, you made the boot work, celebrate it big time! 
I think you can prime the tiny details when you have the time and make it a marvellous piece for your portfolio! And i'm guessing you will have a bigger understanding for what you need to solve in this code now! You CAN get the last function to work! 🥳

PR #149, opened 2023-02-11
Project Chatbot

fannystenberg:


code/script.js
```diff
@@ -29,23 +47,135 @@ const showMessage = (message, sender) => {
...
+    userReply(message)
+    setTimeout(() => happyChoice(message), 1000)
+  } else {
+    userReply(message)
+    setTimeout(thankYou, 1000)
+  }
 }
 
-// Starts here
+// CHAT BOT STARTS HERE // CHAT BOT STARTS HERE // CHAT BOT STARTS HERE //
```
emiliasaberski, line 85: Great to use uppercase letters to show where the different parts of the bot begins

code/script.js
```diff
@@ -29,23 +47,135 @@ const showMessage = (message, sender) => {
...
+    .addEventListener('click', () => nextQuestion('Asian'))
+  document
+    .getElementById('middleEasternBtn')
+    .addEventListener('click', () => nextQuestion('Middle Eastern'))
+}
+// SECOND QUESTION //
+const showOptions = (type) => {
+  questionNr++
+  botReply(`Oh so you're in the mood for ${type}? Yum..🤤`)
+  botReply(`Select a place from the options!`)
```
emiliasaberski, line 122: Maybe use setTimeout here to make the second reply appear a second or so after the first one.
Like this:
```
>       setTimeout(() => {
>     botReply(`Select a place from the options!`)
>     }, 1000);   
```

emiliasaberski (commented):
I think you've done a great chatbot that works exactly as you want it to. Your code is clean and understandable with good commenting. I think it's clear looking at your code that you have the basic understanding of Javascript and how to work with CSS and HTML.

Great job this week! 💪🏻

PR #156, opened 2023-02-12
Chat bot

ninastroyer:
https://creative-custard-60a91d.netlify.app/


code/script.js
```diff
@@ -1,51 +1,177 @@
...
+};
+
+
+//Thank you! Have a great day!!
+const lastMessage = () => {
+    showMessage(`Thank you, have a great day!`, 'bot');
+  inputWrapper.innerHTML =``;
+};
+
+const handleInput = (event) => {
```
edv-rd, line 120: The way your bot is structured, you do not really need this function (handleInput). You're only calling it once, after the greetUser-function, and from then on your other functions are just calling the next one manually.

If you wanted to use this, you should put questionIndex++; in the start of every other function, and make all the click-events in the addEventListeners call handleInput instead. That way, the questionIndex-variable is increasing by 1 for every function it goes through, and the if-loop in handleInput will take care of the rest.

You could get rid of it completely by deleting it and have the addEventListener at line 172 call handleUserName directly.

I am not saying you HAVE to do this as the way you are doing it works fine and is actually easier to read and understand! :)

edv-rd (commented):
In general this is a nice and functional chatbot, it works and it looks good. Your comments are also helpful.

Do you have the Prettier-extension installed in VSCode? Some of your indentation looks a little off, you can see it if you compare activityQuestion and whereQuestion. I suggest installing the extension as it will automatically make your code look pretty.

Great job completing the first project!

PR #159, opened 2023-02-12
Chatbot for week 1

JohLeo:


Nahnahke:
I think the code works well for the purpose of the assignment and fulfils the requirements. It seems the timeout function is not working properly though. Some params without function. 

I think the comments from Johanna are very clear and you can easily follow her code and understand the different functions. 

I think it's a great project and I am impressed by the amount of options she has made for the botReply answers. 

Great job!

PR #161, opened 2023-02-12
Project Chatbot - Frida M. Gustavson

Cheroptera:
This is my chatbot - The Knock Knock Machine :)


code/script.js
```diff
@@ -1,51 +1,215 @@
...
-  }
+  
+}
+
   // This little thing makes the chat scroll to the last message when there are too many to be shown in the chat box
   chat.scrollTop = chat.scrollHeight;
 }
 
+//This function makes the questions and answers appear in the correct order. 
+
```
MalinSkill, line 54: Very impressive structure of the code here, with the else...if statement. Also very clear comment to the function that describe what it does. I did not seem to manage to mark all the lines that I intended to do. But this comment refers to the lines from 53 to 97.

code/style.css
```diff
@@ -15,6 +15,7 @@ h1 {
   line-height: 34px;
   color: #fff;
   text-align: center;
+  font-family:'Gill Sans', 'Gill Sans MT', Calibri, 'Trebuchet MS', sans-serif;
```
MalinSkill, line 18: Great that you have added a generic-family font  for fallback here.

code/script.js
```diff
@@ -1,51 +1,215 @@
...
+userInput.value = '' //Clears input field after submitting
+
+})
+
+//Makes the chat bot move on to the next line.
+sendBtn.addEventListener ('click', () => nextLine(userInput.value)) 
+
+userInput.addEventListener('keypress', (event) => {
+  if (event.key === 'submit' && userInput.value) nextLine(userInput.value)
+  });
```
MalinSkill, line 185: Could this section have been solved with less addEventListener? I don't know the answer to this, but I'm thinking that since the user input will always be in the same text field and submitted the same way. Maybe there is a simpler way write it.

code/script.js
```diff
@@ -1,51 +1,215 @@
...
+    botReply(`Who.`);
+  }
+
+  //Ninth reply from bot 
+  const ninthLine = () => {
+    lineNumber++;  
+    botReply(`What are you? An owl?`);
+    setTimeout(() => botReply (`That's all I have for today.`), 2000) ; 
+    setTimeout(() => botReply (`I hope you enjoyed it!`), 3000) ;
+  }
```
MalinSkill, line 158: Nice timing here with the follow-up botReplys

code/script.js
```diff
@@ -1,51 +1,215 @@
...
 
+let lineNumber = 1; 
+
+const botReply = (msg) => {
+  showMessage (msg, 'bot') ; 
+}
+
+const userReply = (msg) => {
+  showMessage (msg, 'user') ; 
+}
```
MalinSkill, line 17: Nice variables so you don't need to add both parameters each time you invoke the showMessage function.

code/script.js
```diff
@@ -1,51 +1,215 @@
...
+  lineNumber++
+  botReply(`You're welcome!`);
+  setTimeout(() => botReply (`Knock, knock.`), 2000) ; 
+  }
+
+  //Fourth reply from bot 
+  const fourthLine = () => {
+    lineNumber++;  
+    botReply(`Boo.`);
+  }
```
MalinSkill, line 124: In this section the indentation varies a bit. You start with two functions that have no indentation, but on the following functions after that there are indentations. Maybe take a look? In my mind there is no need to intend the functions, since they are on their own. Do you agree?

MalinSkill (commented):
Very fun theme on the chatbot. The conversation flow is very good. I found it easy to read and understand your code, thanks to the comments and nice structure. I have added some notes on small changes in regards of the formatting of the code. The use of indentations and also I noticed sometimes the ; is like this )  ; sometimes );  . I wish I could have come up with more comments, but I guess my skills in reviewing will evolve over time.

PR #169, opened 2023-02-12
Project chatbot

MalinSkill:
Link to my deployed project: [https://dinner-guide.netlify.app]

I was off to a rough start, having big problems understanding the event listeners and how to pass the input from user and make it appear as an answer from user. Once I got the first user reply to appear in the chat, the remaining questions and answers went easier. It was fun to practice javascript doing this chatbot. It was very helpful to look at the code in the example and from earlier students project.  If I had more time I would have tried to organise the code better and look into the sound effects.


code/script.js
```diff
@@ -1,43 +1,214 @@
...
+  showMessage("Welcome, What's your name?", 'bot');
+}
+
+// Initial button click, here I should get the name entered
+sendBtn.addEventListener('click', (event) => {
+  if (nameInput.value === "" || undefined) {
+    showMessage("Sorry I did not catch that, please enter your name!", 'bot')
+    event.preventDefault()
+  } else {
+    event.preventDefault() //this is so the page do not reload.
```
dannebrob: In this event listener you have `event.preventDefault()`  in both blocks of code; maybe move it outside the if...else block? Then you wont need the else statement
Something like this:
```
sendBtn.addEventListener('click', (event) => {
  event.preventDefault()
  if (nameInput.value === "" || undefined) {
   ....
  }
```

code/script.js
```diff
@@ -1,43 +1,214 @@
...
+let userName;
 // Declare your functions after this comment
 
 // This function will add a chat bubble in the correct place based on who the sender is
+//without the ${message} the actual message(parameter) in const greetUser will not be passed, only string message. Only the parameter 'bot' 
 const showMessage = (message, sender) => {
-  // the if statement checks if the sender is 'user' and if that's the case it inserts an html senction inside the chat with the posted message
+  // the if statement checks if the sender is 'user' and if that's the case it inserts an html section inside the chat with the posted message
   if (sender === 'user') {
+    console.log("Now the user is replying");
```
dannebrob: Great use of console.log()! Do you think you need it still in the code for the end users, if not, you might want to remove it / comment it out? 
(also on line 27)

code/script.js
```diff
@@ -1,43 +1,214 @@
...
+}
+
+const userReply2 = (type) => {
+  showMessage(`${type}`, 'user');
+  setTimeout(() => displayRecipeLink(type), 1000);
+}
+
+const displayRecipeLink = (answer) => {
+  if (answer === 'Spicy') {
+    showMessage(`Excellent choice! Find the link to recipe here <a href="https://ohsheglows.com/2020/04/25/cozy-at-home-spicy-any-veggie-soup/" target="_blank">${answer} Soup</a>`, 'bot');
```
dannebrob, line 129: Nice to get a recipe for the suggested dish!

code/script.js
```diff
@@ -1,43 +1,214 @@
...
+  if (reply === 'Yeeees!') {
+    location.reload();
+    return false
+  } else {
+    showMessage('Ok fine, I tried! Maybe you are not even hungry 👻', 'bot')
+    inputWrapper.innerHTML = '';
+  }
 }
 
 // Set up your eventlisteners here
```
dannebrob: Unnecessary comment? (Line 214)

code/style.css
```diff
@@ -147,3 +174,23 @@ button:hover {
...
+  body {
+    background-image: url("./assets/background-medium.jpg");
+  }
+}
+
+@media (min-width: 1024px) {
+  body {
+    background-image: url("./assets/background-large.jpg");
+  }
+}
```
dannebrob, line 196: Nice with responsive images!

dannebrob (commented):
It's an excellent and compelling web app; now I have to go and try my suggested recipes!
Your "if... else" statements are clear and easy to understand. And your comments are helpful and make it easy to know what's going on in the code. 
All the project requirements were met with high standards, and I enjoyed the colour choices you used in the CSS.
Great work, Malin; reviewing your code has been a pleasure. Keep up the good work!

code/script.js
MalinSkill: Thank you for suggestion, I have now moved it to be outside of the if else block. I think I still need to keep the else statement, there is what happens if the name is entered. It might be a bit hard to see with all the comments

```sendBtn.addEventListener('click', (event) => {
  event.preventDefault() 
  if (nameInput.value === "" || undefined) {
    showMessage("Sorry I did not catch that, please enter your name!", 'bot')
    event.preventDefault()
  } else {
    // Store the value in a variable so I can access it after we clear it from the input. 
    // The variable were the value will be stored is outside of this function to be able to use it in more places. In here it get assigned
    userName = nameInput.value
    showMessage(`${userName}`, 'user');
    // Clears the input field
    nameInput.value = ''
    //Here I call the function where I present the dishes to choose from. I will also pass the userName
    setTimeout(() => whatKindOfDish(userName), 1000)
  }
})```

code/script.js
MalinSkill: Good point! I have removed it now

PR #170, opened 2023-02-12
Project Chatbot

camcron:
Link to deployed project:
https://sprightly-phoenix-71884e.netlify.app/

Had a lot of trouble understanding the basics of JavaScript on to get started on the project. 
I had managed to do something wrong with my first fork, therefore there I've only pushed once to my GitHub today.


code/script.js
```diff
@@ -1,51 +1,159 @@
...
+        `<button id="orderBtn">Place order</button>`;
+
+      const orderBtn = document.getElementById('orderBtn');
+        orderBtn.addEventListener('click', (event) => {
+          event.preventDefault();
+        alert("Thank you for your order 🚀");
+        window.location.reload();
+        });
+        }, 3500); 
+      } else {
```
SofiaGerdmar: Good job, getting the 'else' to work! One suggestion would be to have more of the months in the if-statement, and maybe just one in the else. When I tried your bot I got the "Sorry" three times before I could order a magazine. Other than that I can find nothing to improve.

camcron: Thank you for your feedback Sofia! I absolutely agree with your comment and have changed the code accordingly.

PR #173, opened 2023-02-12
Plant bot / Vio

code-and-cats:
https://vermillion-kashata-dd952d.netlify.app/


code/script.js
```diff
@@ -27,25 +26,141 @@ const showMessage = (message, sender) => {
...
+const handleNameInput = (event) => {
+  event.preventDefault();
+  const name = nameInput.value;
+  userMessage(name);
+  
+  if (name !== '') {
+    setTimeout (() => waterQuestion(), 1000);
+  } else {
+    botMessage ("Please fill in a name!");
+  }
```
JosefinRobertsson, line 287:
``` javascript
if (name !== '') {
    setTimeout (() => waterQuestion(), 1000);
  } else {
    botMessage ("Please fill in a name!");
  }
```

I love this! A way more reliable way to handle an empty input field than the "required" attribute in the html-code

JosefinRobertsson:
Hi Vio! When I got to the end of your script.js-file I noticed that from line 151 nothing happens, the last question and last botmessage doesn't show up in the browser. Possibly the placement of the timeout? I will continue the review after you get a chance to check it out :)

JosefinRobertsson:
Hello again! I took another look and the problem is I think that you called the last message botReply and not botMessage. Easy fix! :)

code-and-cats:
> 

Thank you! Fixed it :)

code/script.js
```diff
@@ -27,25 +26,141 @@ const showMessage = (message, sender) => {
...
+  document
+  .getElementById('lotWaterBtn').addEventListener('click', () => {userMessage("I love watering, I keep drowning my plants...");
+  inputWrapper.innerHTML = '';
+  setTimeout(() => waterResponse('lotWater'), 1000)});
+  document.
+  getElementById('naWaterBtn').addEventListener('click', () => {userMessage("I water my plants like a pro!");
+  inputWrapper.innerHTML = '';
+  setTimeout(() => waterResponse('anyWater'), 1000)});
+
+const waterResponse = (waterAnswer) => {
```
JosefinRobertsson, line 314: I like that you made the effort to add this response before the next question. It makes it feel more like you are interacting with the bot and adds to the feeling that the bot wants to be helpful.

code/script.js
```diff
@@ -27,25 +26,141 @@ const showMessage = (message, sender) => {
...
+//Question 1, user clicks, triggers bot answer and question 2
+const waterQuestion = () => {
+  botMessage(`How much do you water your plants, ${nameInput.value}? `);
+  inputWrapper.innerHTML = `
+  <button id="littleWaterBtn">A little</button>
+  <button id="lotWaterBtn">A lot!</button>
+  <button id="naWaterBtn">As needed</button>
+  `
+  document
+  .getElementById('littleWaterBtn').addEventListener('click', () => {userMessage("I keep forgetting to water plants!");
```
JosefinRobertsson, line 302: It's hard to find anything to comment that could be improved upon... I'm thinking here you could have saved a few lines of code and thus improved a tiny bit on readability if you had the timeouts in the user message function. But if you look at it another way, putting them here makes it easy to see what's going on without scrolling back up so, I think there are benefits to both ways.

code/script.js
```diff
@@ -27,25 +26,141 @@ const showMessage = (message, sender) => {
...
+//Event listener for handling the name input
+nameForm.addEventListener('submit', handleNameInput);
+
+//Question 1, user clicks, triggers bot answer and question 2
+const waterQuestion = () => {
+  botMessage(`How much do you water your plants, ${nameInput.value}? `);
+  inputWrapper.innerHTML = `
+  <button id="littleWaterBtn">A little</button>
+  <button id="lotWaterBtn">A lot!</button>
+  <button id="naWaterBtn">As needed</button>
```
JosefinRobertsson, line 299: Was this id meant to be an (as needed) instead of na? I think overall you have an easy to follow-use of params and id's/names, just got a little confused about your choice here

JosefinRobertsson (approved):
Very helpful plant bot! A good idea that I think could be “real life” useful.

I like that the plant suggestion from the bot includes a picture. The style of it adds nicely to the overall styling to!

Great feature that the user can start the bot over if you want to have another go in the end.

I think the choice of font for the title fits the overall feel very well as does the dotted border and it all adds to the style without being gaudy. Less is more.

I know this was not in any requirements, just an idea I had while trying your bot out; it would be neat if the chat bubbles were slightly faded in when they appear. I think it would go well with the soft greens and the plant theme.

Your code is easy to follow and well organized.

I’d love to see this bot with even more content and the possibility to narrow down the best choice of plant for the user even more. It could really be useful and it’s also fun to use. Well done!

code/script.js
code-and-cats: Ah nice catch, it was supposed to be na as in N/A (not applicable) but I think I changed it to "any" in the other question. But missed it here. Thanks!

PR #176, opened 2023-02-12
Quizzie - your swedish quiz-bot

jonsjak:


IrupePozoGraviz (changes requested):
Good work! What i can se you met the requirements and you have a functioning bot that was fun to play and easy to follow. The functions work smoothly and all the buttons work and send me where they should.

One thing that is good to look at is the responsiveness it looks a bit wonky and Im not sure if its mobile first? 

the documentation was good perhaps explain the super short comments a little bit more like on line 164 so if some one else has to do changes they understand what that part means perhaps just write if the user don't want to play or if the user answer no.
otherwise aswesome job!

PR #183, opened 2023-02-12
Chatbot-project

AndreaHedstrom:
Netlify: https://shillyshallyartgallery.netlify.app/


code/script.js
```diff
@@ -1,95 +1,185 @@
...
-const nameInput = document.getElementById('name-input');
-const sendBtn = document.getElementById('send-btn');
+// All the DOM selectors stored as short variables
+const chat = document.getElementById('chat')
+const inputWrapper = document.getElementById('input-wrapper')
+const input = document.getElementById('input')
+const sendBtn = document.getElementById('send')
 
-// If you need any global variables that you can use across different functions, declare them here:
+let botQuestion = 1
```
fikamate, line 9: What does this variable do? (I also have it in my code lol but i have no idea what its doing except that it makes everyting work so i just have it there haha).

fikamate (approved):
Very nice work!!! I did not have much to comment on, as far as where my my own knowledge strikes at least, and that is very much on a basic level.. maybe no need to have a disclaimer on this cause were all learning but you know :) Anyways, what i did was i tested the bot (clicked on all the different buttons in the different categories) and made sure everyting worked the way it should (which it totally did so great job!), then I read through your code and all the commits, i made one comment and i checked it with the breif and teh instructions and as far as i can see youve done all the iterations, except for maybe nr 5, adding a "invalid answer reply (but im not sure how to include that in the current button-functions so that might be a different issue). 

One general input from me would be to maybe try and add more comments thruout the JS-file, documenting what every section does. That would make it easier for others to contribute if it were a team-project, and i think maybe also for yourself when picking up the work later on. At least for me it helps me to understand every step along the way while learning! 

One thing about the code I especially very much appreciate is the commit-workflow, its easy to follow what you’ve done and the logic in your work! Fun to follow how you built up the project step by step!

Also, Its a very fun and nice looking chatbot, and I love that the bot replies rhymes in all the answers 😍

About the images you want to show based on the users choice, really great idea! Im sure you’ll be able to do this in no-time just in a couple of weeks as we learn more :) 

Keep up the good work!! :)

PR #193, opened 2023-08-28
project-chatbot-lauralyckholm

LauraLyckholm:
Link to deployed version:
https://project-chatbot-lauralyckholm.netlify.app/ 

It was really hard do get going with this one! It wasn't until I watched Jenny's live coding on how she made the starter code that I fully grasped the task. When I started adding the event-listener, it forced me to think about what I actually wanted to achieve, and from there I started to google whatever I needed help with.

One problem I had in the beginning was that I was trying to reach the text the user had typed into the input (in this case the users name), but found that very tricky. I tried making the variable that stored the value global, but it still wouldn't log it on testing. The problem was that the answer was logged within a function, and making the variable global didn't work because Javascript loads asynchronously = the submitForm function had not finished executing when I tried to access it. After a while I figured I could just create a new function for showing the answer to the name-input and then calling that within the function submitForm instead - and that worked. 

I got really stuck after the first interaction. So when trying to get going after having created the new buttons for Yes and No, there was a lot of trial and error and testing. I don't really remember what finally made me solve it, but I think it had something to do with me not getting the hang of eventlisteners properly. Once I found the question on StackOverflow regarding how to trigger a new question (https://stackoverflowteams.com/c/technigo/questions/3407) it sort of landed into place.

Once I got further into the code I realized that it all looked very messy. My functions where very long and it was like I walked deeper and deeper into a labyrinth, and I din't know how to get out of it. That's when the reusability of functions actually clicked into place. Attending the Q&A with Diego, where he once more went trough parameters and arguments made that part click for me, and I decided that evening to try and break down my code into smaller chunks. Oh the joy when everything still worked the way it should! I was expecially proud of the createButtons function where I used the parameters in both the text AND the id - killing to flies in one go so to speak. 

The CSS and HTML for the typingIndicator are pretty much copied over from here: https://codesandbox.io/s/chat-typing-indicator-demo-ehneh?file=/index.html
I changed the color to my own, and instead of putting it in the HTML and CSS i created functions and variables for it to create everything dynamically in the script instead.
I though about adding the typingIndicator code into a new scrip for a cleaner look - but maybe next time. 

With more time I would like to add an API to the bot, so that the generated answers are really generated and not hard-coded. I would also like to add a "start over" and "quit"-button at the end instead of prompting the user to shut down the window. I tried creating a restartChat function but it was a bit too tricky and my time ran out.


code/script.js
```diff
@@ -28,24 +86,343 @@ const showMessage = (message, sender) => {
...
+  // This is where the "code" argument is replaced with the actual code
+  insertStyle(`
+    .answer-btn {
+      width: 100%;
+  }`
+  );
+};
+
+/* ------------- Function to create select interface ------------- */
+// The same parameters are used for creating the id's, but since I want these in lowercase, the method toLowerCased is used.
```
IdahCollin, line 130: Really nice and detailed!

code/script.js
```diff
@@ -28,24 +86,343 @@ const showMessage = (message, sender) => {
...
+    list.appendChild(li);
+  }
+  // Scroll the last item into view
+  const lastListItem = list.lastChild;
+  lastListItem.scrollIntoView();
 }
 
-// Set up your eventlisteners here
+/* ------------- END INTERACTION ------------- */
+// An extra message for when the user chooses to end the chat.
```
IdahCollin, line 417: Nice to see a different way to do it instead of asking the user to reload the page!

code/script.js
```diff
@@ -1,15 +1,72 @@
...
+const inputField = document.getElementById('name-input');
+let soundEffect = new Audio('assets/pop.mp3'); // Sound effect to the chat
 
-// If you need any global variables that you can use across different functions, declare them here:
+// GLOBAL VARIABLES
+let userAnswer = ""; // stores users answer in a variable globally for use anywhere
+let selectedGenre = ""; // stores selected genre in a variable globally for use anywhere
 
+// GLOBAL OBJECTS
+// Stores the different movie-choices in an object for easier handling further down.
```
IdahCollin, line 14: Great job using the arrays! Not sure on how and guessing here, but is there a way of making this code (even) more great by merging the oldies and newbie in one array? An array in an array/nested somehow? :)

code/script.js
```diff
@@ -20,6 +77,7 @@ const showMessage = (message, sender) => {
     `
     // the else if statement checks if the sender is a bot and if that's the case it inserts an html senction inside the chat with the posted message
   } else if (sender === 'bot') {
+    soundEffect.play();
```
IdahCollin, line 80: Awesome! Enjoyed the type of sound you picked as well!

code/script.js
```diff
@@ -28,24 +86,343 @@ const showMessage = (message, sender) => {
...
+/* ------------- Function for adding style ------------- */
+/* Creates a function called style, that takes "code" as a parameter. A style-element is created and stored in a variable called "style". The "code" (placeholder for the code - the actual code is added in the argument of an other function) is then added into the style-element by using innerHTML. The style element is then appended into the element "head".*/
+function insertStyle(code) {
+  var style = document.createElement('style');
+  style.innerHTML = code;
+  document.getElementsByTagName("head")[0].appendChild(style);
+};
+
+/* --------------------------------------- */
+// Here I decided to add the "code" into a variable instead of straight into the argument of insertStyle, for a cleaner code since there is a LOT of it :)
```
IdahCollin, line 196: Nice!

code/script.js
```diff
@@ -28,24 +86,343 @@ const showMessage = (message, sender) => {
...
+    showMessage(`Nice to meet you ${userAnswer}! Would you like a movie tip for tonight 😃?`, 'bot');
+    inputField.remove();
+    submitBtn.remove();
+    OptInQuestion();
+  } else {
+    showMessage(`Please enter a name so that I can assist you with a hint of personal touch! 🤓`, 'bot');
+  }
+};
+
+/* ------------- THIRD INTERACTION ------------- */
```
IdahCollin, line 281: Your comments throughout the file really helped me during the review :)

code/script.js
```diff
@@ -28,24 +86,343 @@ const showMessage = (message, sender) => {
...
+  }
+  selectDropDown.addEventListener("change", checkValue); // on the change of a select option, the code for checkValue is invoked.
+
+};
+
+/* ------------- FIFTH INTERACTION ------------- */
+// Lets user select from an old movie or a new movie
+function oldieOrNewbie() {
+  const oldie = document.getElementById("btn-one");
+  const newbie = document.getElementById("btn-two");
```
IdahCollin, line 345: What do you think about changing this function to the one with arrows to use the same "method" throughout the js? 
```
const oldieOrNewbie = () => {
  const oldie = document.getElementById("btn-one");
  const newbie = document.getElementById("btn-two");
```

IdahCollin (commented):
GREAT job Laurawesome! Your code is working, fulfills requirement (and then some), follows guidelines and is well described. Added one or two tiny comments. Wow, giving we're at week 2 in the boot camp this is great! I learned more code by reviewing your work. /Idah

code/script.js
LauraLyckholm: You're right!! That would probably make much more sense.. Perhaps I was a bit rusty on working with objects at this point, or just tired over all 😅 Thanks for the tip!

code/script.js
LauraLyckholm: Perhaps you're right! Sometimes I worked with the function declaration for being able to use the functions where ever in the code, but in this case I might be able to use an arrow function too. It was however also nice to test out doing it in different ways 🤭

PR #197, opened 2023-08-31
project-chatbot-Mina

Minadarabi:
Deployed on Netlify : https://jade-jalebi-01e7d3.netlify.app/

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -297,6 +297,11 @@ let totalPrice = 0;
   });
     
     sendBtn.addEventListener("click" , ()=>{
+      if(counter === 0){
+        sendBtn.disabled = true;
+        alert("Oops! You have to choose at least one.");
+        return
+      }
       showMessage(`${counter}` , 'user');
       addBtns.style.display = "none";
       
@@ -355,7 +360,7 @@ let totalPrice = 0;
       childBtn.addEventListener("click" , ()=>{
         showMessage("Child size" , 'user');
         sizeBtns.style.display = "none";
-        let childTotalPrice = totalPrice - 20
+        let childTotalPrice = totalPrice - 20;
         setTimeout(()=>{
           showMessage(`Your order will be ${counter}-${selectedFood} Child size and the price would be ${childTotalPrice} SEK ` , 'bot')
           showConfirm()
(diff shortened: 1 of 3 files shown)


code/script.js
```diff
@@ -1,14 +1,68 @@
...
+
+btnsEl.style.display = "none";
+nameFormEl.style.display = "flex";
+pizzaMenue.style.display = "none";
+pastaMenue.style.display = "none";
+saladMenue.style.display = "none";
+addBtns.style.display = "none";
+sideBtns.style.display = "none";
+
+const pizzaPrice = {
```
Shiqwan, line 44: nice to add prices! And then calculating the prize!

code/script.js
```diff
@@ -29,23 +83,358 @@ const showMessage = (message, sender) => {
...
+  e.preventDefault();
+
+  const enterName = nameInput.value.trim();
+  
+  if(enterName !== ""){
+    message = nameInput.value;
+    showMessage(enterName , 'user');
+    clearInput();
+    setTimeout(() => {
+      showMessage(`Nice to meet you ${enterName}.what type of food would you like to order?` , 'bot')
```
Shiqwan, line 108: You're so focused on the code (which I understand, and you should be!), if you spent two minutes more on the statements, it would be flawless! It's easy corrections that you know, you're so careful with the code, it feels sloppy when the text is not as well done! You have however focused on the important things when it comes to the project!

code/script.js
```diff
@@ -29,23 +83,358 @@ const showMessage = (message, sender) => {
...
+
+    if(valueOptionPizza === "Margarita"){
+      showMessage("Margarita" , 'user');
+      setTimeout(() =>{
+        showMessage("please choose how many Margarita Pizaa would you like?" , 'bot');
+        addBtns.style.display = "flex";
+        selectedFood = "Margarita Pizaa";
+        price = pizzaPrice.Margarita
+
+      },1000)
```
Shiqwan, line 150: This is so interesting, adding the 1000 at the end, never would have thought, cool that it works like that!

Shiqwan (approved):
I don't have a lot of comments because you've done some things I don't know what they are! But that's a good thing, you've put so much work into this, it shows! If you take a second on the things that the "customer" sees for your next project, you'll get even better!
I would also enjoy comments if you could find the time(I get that it's hard to find the time) but that's of course up to you, if that would be helpful to you and not just the coder that might follow you! Overall, great work Mina, gold star!

code/script.js
Minadarabi: Thank you for your review and comment Ulrika

PR #200, opened 2023-09-01
My bot-project!

SVMaxica:
This was a really hard project, I feel like I understand the code when I read it now but if I look at this project in a few week I will probably need some time to understand it. I need a lot more practice a this.
https://bobtherestaurantbot.netlify.app/

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,4 +8,4 @@ I started to plan my questions from the bot and then i built the question one by
 
 ## View it live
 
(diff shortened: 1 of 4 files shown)


FridafridaL (commented):
Nice work Sandra! I like how you structured your code, it's easy to follow and relatable to the chatbot in the browser. 
I didn't know you could just use an (e) instead of (event) in this 

nameForm.addEventListener('submit', (e) => {
  e.preventDefault();
  
  on line 44 so that was nice to see.
  
  You did a great job and the chatbot works fine! Hurray!!

PR #203, opened 2023-09-01
Pharmacy Evaluation-Chat Bot

isarobertini:
Link to project:
https://pharmacy-evaluation-chat-bot.netlify.app/


code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+    in order for us to improve our service.`, 'bot');
+  pop.play();
+  startBtn.style.display = 'none';
+
+  setTimeout(() => {
+    showMessage(` First of all, what's your name?`, 'bot');
+    pop.play();
+  }, 4000);
+
 }
```
AmandaRam: Preferably all of this would be in one message to provide a more pleasant user experience

code/script.js
```diff
@@ -1,14 +1,25 @@
...
+const ageBtn1 = document.getElementById('ageBtn1');
+const ageBtn2 = document.getElementById('ageBtn2');
+const yesBtn = document.getElementById('yesBtn');
+const noBtn = document.getElementById('noBtn');
+const dropDown = document.getElementById('dropDown');
+const notGoodBtn = document.getElementById('notGoodBtn');
+const okayBtn = document.getElementById('okayBtn');
+const goodBtn = document.getElementById('goodBtn');
+const startBtn = document.getElementById('startBtn');
+var pop;
```
AmandaRam, line 16: Use const instead of var.

code/script.js
```diff
@@ -1,14 +1,25 @@
...
+pop = new Audio('pop.mp3');
```
AmandaRam, line 17: Unnecessary redeclaration of variable.

code/script.js
```diff
@@ -1,14 +1,25 @@
...
 
 // This function will add a chat bubble in the correct place based on who the sender is
 const showMessage = (message, sender) => {
-  // the if statement checks if the sender is 'user' and if that's the case it inserts an html senction inside the chat with the posted message
+  console.log("MESSAGE IS:", message); //console.log messages 
+  console.log("SENDER IS:", sender); //console.log messages
```
AmandaRam: Unnecessary console.logs.

code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+  // showMessage function is being called, greets the user.
+  showMessage(`Hello there, welcome to the evaluation sheet! 
+    We are really happy that you take your time to answer some questions
+    in order for us to improve our service.`, 'bot');
+  pop.play();
+  startBtn.style.display = 'none';
+
+  setTimeout(() => {
+    showMessage(` First of all, what's your name?`, 'bot');
+    pop.play();
```
AmandaRam: The function that plays the sound can be inside the showMessage function so you don´t have to repeat yourself.

code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+  }, 4000);
```
AmandaRam, line 59: You can declare this timeout (e.g 4000) as a global constant variable to not repeat yourself. Then if one would like to change the delay, one would only have to change it in one place instead of several places :)

code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+
 }
 
-// Set up your eventlisteners here
+// Function storing the name of the user.
+const handleNameInput = (event) => {
+  event.preventDefault();
+
+  const name = nameInput.value;
+  showMessage(`${name}`, 'user');
```
AmandaRam, line 81: Can be written as `showMessage(name, 'user')`.

code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+        handleageBtn(event, name, 'under 18'); //parameter string under 18 passed as an argument to ageGroup
+        ageBtn1.style.display = 'none';
+        ageBtn2.style.display = 'none';
+        inputWrapper.style.display = 'block';
+      };
+      ageBtn2.onclick = (event) => {
+        handleageBtn(event, name, 'over 18');//paramter string over 18 passed as an argument to ageGroup
+        ageBtn1.style.display = 'none';
+        ageBtn2.style.display = 'none';
+        inputWrapper.style.display = 'block';
```
AmandaRam:
```
ageBtn1.style.display = 'none';
ageBtn2.style.display = 'none';
inputWrapper.style.display = 'block';
```

is duplicated and can be moved inside the handleageBtn function.

code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+        showMessage(`${response}!`, 'user');
+        pop.play();
+        if (response === `yes`) {
+          showMessage(`Amazing!`, 'bot');
+          pop.play();
+        } else {
+          showMessage(`You need to think about your age. This page will be automatically reloaded.`, 'bot');
+          pop.play();
+          yesBtn.style.display = 'none';
+          noBtn.style.display = 'none';
```
AmandaRam: The buttons could be put in a div in the HTML so that one would only have to hide the div instead of two buttons.

code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+            goodBtn.style.display = 'block';
+
+            notGoodBtn.onclick = () => handleReview('Not good', name);
+            okayBtn.onclick = () => handleReview('Okay', name);
+            goodBtn.onclick = () => handleReview('Good', name);
+
+
+          }, 2000);
+
+        }, 2000);
```
AmandaRam: These setTimeout functions can be separated to get a cleaner code. Otherwise clear and easy to read :)

code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+            }, 5000);
+
+          }
+        }, 1000);
+      };
+
+
+
+    }, 3000);
+  }, 1000);
```
AmandaRam: It seems like the whole JS file is wrapped inside of setTimeout, which makes the code hard to understand at times. Preferably all functions should be declared using only a few rows.

code/script.js
```diff
@@ -17,35 +28,219 @@ const showMessage = (message, sender) => {
...
+
+};
+// Event listener, when user press enter.
+nameInput.addEventListener('keydown', (event) => {
+  if (event.key === 'Enter') {
+    handleNameInput(event);
+  }
+});
```
AmandaRam, line 234: Bind the function to the submit event instead because now the bot only works if you submit your name using the Enter key. If you press the send button it submits the form without preventing the default behaviour (e.g. reloading the page).

AmandaRam (changes requested):
The bot only works in some instances. Nested setTimeouts lead to unpredictable timings between messages. Also the code should be less repetitive. 

Overall it works fine and I really like the twist of the design and animations! Good job! :)

PR #210, opened 2023-09-03
Week 2 - Project Chatbot - submitted

BeckieMorton:


code/script.js
```diff
@@ -1,25 +1,36 @@
...
 
-// This function will add a chat bubble in the correct place based on who the sender is
+//----------This function will add a chat bubble in the correct place based on who the sender is-----//
 const showMessage = (message, sender) => {
-  // the if statement checks if the sender is 'user' and if that's the case it inserts an html senction inside the chat with the posted message
+  // the if statement checks if the sender is 'user' and if that's the case it inserts an html section inside the chat with the posted message
   if (sender === 'user') {
+    console.log(sender); //Step 1 - added to check value
+    userAnswered = true;
+    console.log(userAnswered); //Step 1 - added to check value
```
mirelcac, line 21: Looking at the code, could you elaborate on the use of true and false?

code/script.js
```diff
@@ -29,23 +40,247 @@ const showMessage = (message, sender) => {
...
+  showMessage(`Type YES or NO`, 'bot');
+  //event listener for this question is logSubmit
+}
+
+
+//---------------- Bot Question 3 - Choose destination continent -------------------------//
+
+const botQuestion3 = (yesOrNo) => {
+  console.log("You have entered Question 3 function"); //reference for me to help me track the flow
+  let userEntersYN = usersName.toLowerCase(); //changes user input into all lowercase so it can be verified
```
mirelcac, line 74: Nicely solved!

code/script.js
```diff
@@ -29,23 +40,247 @@ const showMessage = (message, sender) => {
...
+
+//---------------- Bot Question 5 - Date to travel -------------------------//
+
+const botQuestion5 = () => {
+  console.log("You have entered Question 5 function"); //reference for me to help me track the flow
+  showMessage(`We have some amazing locations in this area`, 'bot');
+  showMessage(`Please use the date picker on the right to choose the date you wish to travel`, 'bot');
+
+  //create date input field to choose date
+  const chooseDate = document.getElementById("input-wrapper")
```
mirelcac, line 188: The calendar was a really cool feature!

code/script.js
```diff
@@ -29,23 +40,247 @@ const showMessage = (message, sender) => {
...
+  }
+}
+
+//when form is submitted
+const form = document.getElementById("name-form");
+form.addEventListener("submit", logSubmit);
+
+//----- First function - When the website is loaded, the chatbot asks first question, after a 1 second delay ------//
+
+setTimeout(botQuestion1, 1000) //Step 1 - played around with the value
```
mirelcac, line 285: I moved this one up to be underneath the message - unclear to me if there is a reason to have it in the bottom as the instructions said. 

Also for event listeners, I had all of them within functions - not sure why they should be outside? I understood the instrutions differently - instead of having all functions in the top, event listeners in middle, etc i understood it as the order to build up a function:
write your function 
event listener for funcion
set timeout for next function
write your next function

I now see how we understood it differently, and I am not sure which way is the "best" or "cleanest". I hope you understand what I mean!

code/script.js
```diff
@@ -29,23 +40,247 @@ const showMessage = (message, sender) => {
...
+}
+
+
+//---------------- Bot Question 6 - Confirmation to proceed -------------------------//
+
+const botQuestion6 = () => {
+  console.log("You have entered Question 6 function"); //reference for me to help me track the flow
+  showMessage(`Our team now have enough information to prepare some options for you`, 'bot');
+  showMessage(`We will leave you with a reference number that will help you access more details for your journey`, 'bot');
+  setTimeout(showMessage(`Would you like as to procceed?`, 'bot'), 1000); //this setTimeout doesn't seem to work?
```
mirelcac, line 210: This is the only place where "showMessage" is within the setTimeout. Otherwise, we link to the next function which contains the message. Maybe that is why?

code/script.js
```diff
@@ -1,25 +1,36 @@
 // Variables that point to selected DOM elements
-const chat = document.getElementById('chat')
+const chat = document.getElementById('chat');
```
mirelcac, line 2: I havent fully gotten the hang of the DOM. I added five links here - send-btn, name-form etc.. but your code seems to work just fine without having them listed up here

mirelcac (approved):
The background picture is cool. You have used many different elements such as buttons, dropdown, input field and calendar, which is probably more of a challenge, but the result is great! The comments are easy to follow as well. Great job!

PR #214, opened 2023-09-03
Project Chatbot

majoh23:


code/script.js
```diff
@@ -36,12 +41,103 @@ const showMessage = (message, sender) => {
...
+  // Bot reply based on user choice
+    switch (userChoice) {
+      case '1':
+        botReply = `Great! I have booked a one-way ticket for you to Samarkand, Uzbekistan. Total price €1439. The details will be sent via email. Have a nice trip! ✈️`;
+        break;
+      case '2':
+        botReply = `Good choice! A single room at City Inn Lodge in Botswana awaits you. I have booked 14 nights with breakfast included. Total price €2279. I wish you a pleasant stay! 🏨`;
+        break;
+      case '3':
+        botReply = `Perfect! A Ford Escape (or similar) is waiting for you in Cheyenne, ID. 9 days rental including free miles and insurance. Total price $1203. Get ready for some inspired driving! 🚗`;
```
KroLuna, line 94: The price is in dollars here and in euros above maybe it could be good to have only one currency.

code/script.js
```diff
@@ -36,12 +41,103 @@ const showMessage = (message, sender) => {
...
+      default:
+        botReply = `Invalid choice. Please select a number from 1 to 3.`;
+    }
+
+  // Bot reply after a delay
+    setTimeout(() => {
+      showMessage(botReply, 'bot');
+    }, 6000);
+
+  // Closing message and additional messages
```
KroLuna: Are this three messages meant to come up together or is user supposed to chose something and then the one of the messages is supposed to pop up?

code/script.js
```diff
@@ -36,12 +41,103 @@ const showMessage = (message, sender) => {
...
+  // Display a welcome message from the bot after a short delay
+  setTimeout(() => {
+    const botMessage = `Hello ${userName}! Let's see the world!`;
+    showMessage(botMessage, 'bot');
+  }, 500);
+
+  // User options after a delay
+  setTimeout(() => {
+    const botMessage = `This is what I can do for you: 
+<p>1 - Book a flight</p>
```
KroLuna: For example you could move this "p's" so they are underneath botMessage, I hope you understand what I'm trying to say 🙈

KroLuna (commented):
I think the code is good and it follows the requirements and it looks good too. I understand your explanations and it would be good to make some order between the lines, like tidied up a little bit. When the confirmation comes, there is a few messages that come up, is it supposed to be like that? Maybe it could be a good idea to take a look at it. Good job!

PR #216, opened 2023-09-03
Chatbot project- sprint 1 week 2

Sheryll-Ann:
https://hotel-lebanian-chatbot.netlify.app/


code/style.css
```diff
@@ -6,14 +6,19 @@ body {
...
   font-family: 'Montserrat', sans-serif;
-  background: #0026ff;
+  background: #235784;
+  background-image: url("./assets/nattu-adnan-atSUvc1hMwk-unsplash.jpg");
+  background-repeat: repeat;
+  background-position: center;
+  background-size: cover;
 }
 
 h1 {
```
VAstrom, line 16: I really love the styling of the h1 element! The shadow looks like a reflection in the water! Super nice!

code/script.js
```diff
@@ -1,51 +1,452 @@
...
+  <button id="boathouse" value="boathouse">Boathouse</button>
+  <button id="diner" value="diner">Diner</button>
+  `
+  //Using addEventListener method on click event to call next function. "CURRYING" in programming lang.
+  document.getElementById('spa').addEventListener('click', activityConfirmation);
+  document.getElementById('golf').addEventListener('click', activityConfirmation);
+  document.getElementById('boathouse').addEventListener('click', activityConfirmation);
+  document.getElementById('diner').addEventListener('click', activityConfirmation);
+
+  // old code saved- to understand the refactoring/currying
```
VAstrom, line 105: I like that you have a lot of comments and also included old code so it's possible to see the progress.

code/script.js
```diff
@@ -1,51 +1,452 @@
...
+  showMessage(`${timeSlot}`, "user");
+  setTimeout(() => { showMessage(`Thank you ${userName}, we have booked your activity "${bookedActivity}" for ${timeSlot}`, "bot"); }, 1000)
+  setTimeout(() => { showMessage(`A confirmation message will be sent to you on your device through our downloaded hotel app.`, "bot"); }, 2000)
+  setTimeout(() => { showMessage(`Enjoy your stay at Le Banian!`, "bot"); }, 3000);
+  setTimeout(restartingPage, 4000);
+}
+
+//Iteration 8
+//End of program, suggest user to reload the page
+const restartingPage = () => {
```
VAstrom, line 435: I like this! A restart button is really smart in case the customer wants to book something else as well!

README.md
```diff
@@ -1,13 +1,72 @@
...
+
+* The
+```
+setTimeout (() => {},1000)
+``` 
+function used for better user experience as all messages appeared in blocks not visible to the eye as the user's sight is focused on changing buttons when interacting with buttons or select, the chat on user's side should prompt the answer simultaneously and the same answer selected or else weird feeling when other words added pops up (like the mind needs to process a new information on the msg).
+
+
+
+* If I had more time,
```
VAstrom, line 64: I like the ideas you have for possible future updates. I agree, I also think it would be really nice if the customer could combine activities like a designing their own bundle of activities. 
I think the change of background could look super nice as well!

code/script.js
```diff
@@ -1,51 +1,452 @@
...
+  showMessage(`${userName}`, "user");
+  nameInput.value = ""; //clear nameInput's value 
+  setTimeout(proposeActivity, 1000);
+}
+
+
+//Iteration 3
+//part 1
+//bot replies back using message proposing activities
+const proposeActivity = () => {
```
VAstrom, line 78: I like the way you have paced the pop up of messages and the buttons/drop-down menus overall. i think that creates a nice user experience. For the name input, I can add my name a second time before the buttons appear since it stays a bit longer but I'm not sure if removing them earlier would be better either.

code/script.js
```diff
@@ -1,51 +1,452 @@
...
+}
+
+//Iteration 5
+//part 1
+//If user confirms (answer is yes)
+const confirmationAnswer = (event) => {
+  event.preventDefault();
+  const answer = event.target.value;
+
+  //Ternary not functioning when answer is no, calling activitySelection oavsett -using if else instead, old code:
```
VAstrom, line 141: Nice that you included this even though it wasn't working as expected. I'm not sure either why it's not working when looking through it.

VAstrom (approved):
Amazing job! The chatbot worked as intended and I can't see any visible bugs! The CSS styling also looks really great! The size of the project is really big with all the options and their respective sub-choices! I imagine that must have taken time to set-up. Well done!
The code is well structured and clear.  It's good that you added explanations in the comments since some code blocks were quite long and the comments made it easier to follow along. 

Since we're learning about objects and arrays this week, something I thought about is that maybe it's possible to store and organize all info about the different activities in an object with arrays or something like that. For instance activities = [
spaActivities : ["Foot massage", "Reiki", "Oil massage"],
golfActivities : [... and so on...
or and array with objects depending on the structure of the info you have and what you would want to include.

PR #218, opened 2023-09-03
malin-lunde-project-chatbot

malinlunde:
I will make pull request and hand in my work but since it doesnt work I will return to it later and make updates.


code/script.js
```diff
@@ -35,17 +42,52 @@ const showMessage = (message, sender) => {
 
 // Starts here
 const greetUser = () => {
-  // here we call the function showMessage, that we declared earlier with the argument "Hello there, What's your name?" for message, and the argument "bot" for sender
-  showMessage("Hello there, What's your name?", 'bot')
-  // Just to check it out, change 'bot' to 'user' here 👆
-}
 
+  showMessage("Welcome to report error messages about our municipal streets! Do you want to make a report?", 'bot');
+};
```
Hang-Nguyen-Vu, line 130: This function has been defined, but not invoked anywhere. That is why the chatbot does not show I guess. You might have accidentally removed it from this file. You could use `setTimeout(greetUser, 1000)` to make the first message appear one second after the website is loaded.

code/script.js
```diff
@@ -35,17 +42,52 @@ const showMessage = (message, sender) => {
...
+
+
+
+
+
+const handleUserAnswer = event => {
+  event.preventDefault ();
+  const userAnswer = userInput.value.trim().toLowerCase();
+  showMessage(userAnswer, 'user');
+  setTimeout(() => {
```
Hang-Nguyen-Vu: After showing the message from the user, you can clear the value from the input field by writing `userInput.value = ''`.

code/script.js
```diff
@@ -35,17 +42,52 @@ const showMessage = (message, sender) => {
...
+
+  if (userAnswer === 'yes') {
```
Hang-Nguyen-Vu: Here you can create two buttons for yes and no and add event listeners to them(click --> something happens), so you can avoid the invalid answers. Just remember to remove the input field, send button and label if you are going to use buttons.

code/script.js
```diff
@@ -35,17 +42,52 @@ const showMessage = (message, sender) => {
...
+    showMessage("Svar", 'bot');
+    
+  } else if (userAnswer === 'no') {
+    showMessage("OK!", 'bot');
+
+  } else { showMessage("I'm sorry, I didn't understand your response. Please answer with 'yes' or 'no'.", 'bot');
+  }
+  reply();
```
Hang-Nguyen-Vu: A suggestion for the next step is to define each function where you put in not only the bot reply to user's previous answer "thank you..." but also the next question you would like it to ask. It could include the following parts: 
1. Bot reply to user's previous answer 
2. Bot ask the next question (can be put together with part 1)
3. Add the intended question and its alternatives in innerHTML, for example buttons, dropdown list etc
4. Store the user's choice in variable(s)
5. Add event listener to user's choice of alternative(for button, we often listen to event "click"; for dropdown list, it's the event "change" --> something happens)
6. Remove the buttons/dropdown list etc 
7. Invoke the next question function with a delay to move on to the next one

Hang-Nguyen-Vu (commented):
Hi Malin! I left a few comments and hope that they can help you finish the project.

PR #219, opened 2023-09-03
CaroL project-chatbot

KroLuna:


Isaackamran (commented):
The coding looks good. For my own navigation, I like to add comments for almost each "new" function so that I know where and what it is.

KroLuna (commented):
I made changes on the name input, if the user types a number the bot will answer saying that they need to type a name with letters and not numbers.

PR #222, opened 2023-09-03
ChatBot-FridaLindskog

FridafridaL:
deployd lin: https://meek-nougat-b116c9.netlify.app/

github: https://github.com/FridafridaL/project-chatbot


klaudiaW0410:
Congratulations on completing the task. I know it wasn't easy. In your code you have included buttons as well as select options. The code looks clean and reads well.

code/script.js
```diff
@@ -1,51 +1,195 @@
...
+  option3Button.addEventListener("click", () => {
+    handleButtonClick("Animal 3");
+  });
+  showMessage(
+    "Which one of these unusual animals would you like to learn more about?",
+    "bot"
+  );
+
+  chat.appendChild(animalButtons);
+};
```
klaudiaW0410, line 92: In this code, to have the buttons in the text input place, you can create a new variable with the id `input-wrapper` and then you can add this variable using innerHTM and instead of the buttons in the form, you have buttons in the input place

code/script.js
```diff
@@ -1,51 +1,195 @@
...
+      // Handle user's "Yes" choice
+      showMessage("Okey, sure!", "bot");
+
+      const dropdownHTML = `
+      <select id="animal-dropdown">
+        <option value=""></option>
+        <option value="Platypus">Platypus</option>
+        <option value="Mantis Shrimp">Mantis Shrimp</option>
+      </select>
+    `;
```
klaudiaW0410, line 149: Here you can use the smel like with button . And Also maby in first option write somethin like --Select animal-- then its lokk more easer for user :)

PR #230, opened 2023-09-03
elbines plant bot

elbines:


code/script.js
```diff
@@ -1,51 +1,252 @@
...
-  showMessage("Hello there, What's your name?", 'bot')
-  // Just to check it out, change 'bot' to 'user' here 👆
+  showMessage(`Hello there!`, 'bot')
+  showMessage(`I'm Anna, a chatbot.`, 'bot')
+  showMessage(`What's your name?`, 'bot')
+}
+
+const userNameInput = event => {
+  event.preventDefault();
+  //I think the name is saved for later, but is it saved globally?
```
sansan-sakura: I think Yes. When you declare a variable in the global context, then it will be globally available.

code/script.js
```diff
@@ -1,51 +1,252 @@
...
+  userName = nameInput.value;
+  nameInput.value = "";
+  showMessage(` ${userName}`, "user");
+  setTimeout(() => reply(userName), 500);
+  setTimeout(() => askForChoise(), 1200);
```
sansan-sakura: I didn't know how you could write the arrow function in your setTimeout. It is good to know!! 
I used setTimeout(reply, 500, userName).

code/script.js
```diff
@@ -1,51 +1,252 @@
...
+  showMessage(`Welcome to 'Elbines-Plants' ${userName}!`, 'bot')
+}
+
+//I wanted a delay between the messages, thats why they are in two functions
+const askForChoise = (YesNo) => {
+  //Do I need to type anything in ()?
+  showMessage(`We can help you by selecting some options for you based on your preferences. <br><br> Would you like to continue?`, 'bot')
+  //Buttons Yes No  will appear
+
+  inputWrapper.innerHTML = //(When) shoud i use value (or class) instead of ID?
```
sansan-sakura: I think it depends on which HTML element you use. The first one is about if you need a value for a select element. And the second link is about how value attribute works differently with different HTML elements.
Here is a link you might be interested in: https://stackoverflow.com/questions/38854683/is-a-value-attribute-required-for-a-html-select-element
This could be interesting too: https://www.geeksforgeeks.org/html-value-attribute/

code/script.js
```diff
@@ -1,51 +1,252 @@
...
+
+
+//Bot's welcome message with username pops up. Asks user to make a choise.
+function reply(firstReply) {
+  showMessage(`Welcome to 'Elbines-Plants' ${userName}!`, 'bot')
+}
+
+//I wanted a delay between the messages, thats why they are in two functions
+const askForChoise = (YesNo) => {
+  //Do I need to type anything in ()?
```
sansan-sakura: You are talking about function askForChoise =(yesNo)=>{} ?? 
If you need more parameters, then you can add as much as you need. But if you don't need it, then you don't need any.

code/script.js
```diff
@@ -1,51 +1,252 @@
...
+
+
+  if (choice === 'Yes') {
+    showMessage(`Please select a catagory from below:`, 'bot')
+    const plantCategory = ['Green Plants', 'Flowers', 'Plant Supplies']
+    showOptionsPlantCat(plantCategory)
+  } else if (choice === 'No') {
+    showMessage(`Ok, I understand. Feel free to browse our webpage. If you need help be sure to ask!`, 'bot')
+  }
+  chat.scrollTop = chat.scrollHeight
```
sansan-sakura: This is a good trick!! I didn't know it, and it seems very useful!!

code/script.js
```diff
@@ -1,51 +1,252 @@
...
+  <option value="PriceRange4">1000,- and up </option>
+  </select>
+  `;
+
+  select = document.getElementById("price-select")
+
+  const priceSelected = document.getElementById('price-select')
+    //Adding eventlisteners to option range 1
+
+    .addEventListener('change', (event) => {
```
sansan-sakura: You might want to add "priceSelected" to the event handler.?

sansan-sakura (commented):
Overall, your code is easy to follow as it is written in order. You wrote many comments to make it easier to understand what is going on your code. Good job!!

I might think you could split your functions into smaller functions for reusability , but I know that is a challenge, and I struggle with it too...

This is an idea I came up with:

function sendMessageWithTimer (message, sender,time){
setTimeout(()=>showMessage(message),sender),time)}

--------------------------------
When I want to call it, I could use like this. I used line 171 from your script.js.
 sendMessageWithTimer("Great! I will select some great options (in the price range ${selectedPriRan}) for you. I hope you love them as much as I do!","bot",700)

PR #235, opened 2023-09-05
Nosy Chat-bot

BreannaSand:


BreannaSand:
Evelyn's original comments from the old pull request which got deleted while trying to fix the bug she notified me of. Her comments:
1. Hi Breanna! 💫 Great job at finishing your chatbot! I think it was a good structure and easy logic to understand. It is so fun to see that you can do in so much different ways, I did not think of the way to create the buttons in the html files. There was a minor bug in the beginning, I did not see the first question from the bot, could it be it needs to have a closing tag? 
2. So smart to create the buttons in the html file! Did not think of that :)
3. I do not see this showMessage on the deployed version, could it be that you forgot the closing tag ; ?
4. The logic is easy to understand! 😊
5. I keep getting this in the console of your chatbot, i dont know that it is?
Attachment-1

EvelynDelCarmen (approved):
Great job Breanna! I did the first review and noticed a bug in the project, we communicated in Slack and Breanna fixed the bug. It all looks great!

PR #236, opened 2023-09-08
Chatbot update

saradahlstrom:


saradahlstrom:
Trying to add more if-statements and imrove the code

PR #240, opened 2023-12-27
Remove some code

HIPPIEKICK:


dzc1 (approved):
Nice cleannup of ghe logic, I think students will benefit from this and nnot be that lost !

PR #275, opened 2024-02-23
Chatbot

linaroos:
## Netlify link
https://65d32863583cf500081d9183--penguin-chatbot.netlify.app/


.DS_Store
NellaMori, line 1: .DS_Store is a file that stores custom attributes of its containing folder, such as folder view options, icon positions, and other visual information, and we don't need to share it in our project. Please add .DS_Store to your .gitignore file (if you don't have one, you need to create it) 👍

AntonellaMorittu (approved):
Hi Lina, good job with your chatbot! 🤖 The code is well structured, you meet the requirements and you also put into practice the skills learnt so far, amazing! ⭐  To improve even more the UX you could add validation for the name field so that is not empty for example? I left a few comments in the code, keep up the good work! 👍

PR #279, opened 2024-09-13
Pink Magic 8 Ball

nella-x:
## Netlify link
Add your Netlify link here.
PS. Don't forget to add it in your readme as well.

https://pink-magic-8-ball.netlify.app/ <----


code/script.js
```diff
@@ -1,5 +1,11 @@
+console.log(`let's chat`)
+
 // DOM selectors (variables that point to selected DOM elements) goes here 👇
-const chat = document.getElementById('chat')
+const chat = document.getElementById('chat');
+const nameForm
+  = document.getElementById('name-form');
```
HIPPIEKICK, line 6: Slight error here (unnecessary newline):
`const nameForm = document.getElementById('name-form');`

HIPPIEKICK (approved):
What a fun idea for a chatbot! 🎱

**HTML/CSS**
- Well structured and styled, not much to say ⭐
- Since you're using the name form for all the inputs, consider changing the class and ID for this (and the corresponding DOM selector) to make it more generic

**JavaScript**
- You've broken down the code into different functions like saveNameAndRespond, and generateMagic8BallResponse. This modular approach is great for readability and reuse. Kudos for descriptive variable naming!
- Easy to follow the flow, and it feels like you're not overcomplicating things.
- Consider disabling the button or in some other way validating the user input so that it's not possible to send empty inputs

**Clean Code**
- Be consistent with whether you're using arrow functions or function declarations.
- Remove console.logs before handing it, they are good for developing but not needed in production.

Nice job, I had fun trying this out 😄

PR #281, opened 2024-09-14
Xings chatbot about travel adventure

xingyin2024:
## Netlify link
https://travelchatbot.netlify.app


HIPPIEKICK (approved):
Nice job with your travel chatbot ✈️ Let's have a look at some feedback!

**HTML/CSS**
- Clean structure and nicely styled 👍 Just be mindful when using text overlay on images, the text is barely readable

**JavaScript**
- Various functions handle different aspects of the conversation, such as showing messages, asking questions, and processing user inputs. This modular approach is excellent for readability and reusability ⭐ 
- Nice with some sound, makes it more interactive! Consider breaking it out into its own function
- LOVE the renderOptions function! Way to make it DRY 🥳 
- Consider adding some validation so that the user can't submit an empty name input

**Clean Code**
- Your descriptive function names really helps with the readability, keep going like this! ⭐ 
- Remember curly brackets for if/else statements, it makes the code easier to read

Keep up the good work Xing, looking good!

PR #286, opened 2024-09-15
Jonas chatbot

Jonash189:
https://jonas-chatbot.netlify.app/

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -40,7 +40,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
   // Greet the user
   const greetUser = () => {
-    showMessage("Welcome to Jonas Cafe! What's your name?", 'bot');
+    showMessage("Welcome to J.Cafe! What's your name?", 'bot');
   };
 
   // Handle the name input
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -1,53 +1,158 @@
...
+    isChoosingDish = false; // Stop asking for dishes after valid choice
+  };
+
+
+  const handleUserInput = (event) => {
+    event.preventDefault();
+    const input = nameInput.value; // Get input from user
+    nameInput.value = ""; // Clear input field
+
+    if (!isAskingForDrink && !isAskingForFood && !isChoosingDish) {
```
HIPPIEKICK, line 138: Hehe, couldn't this be isAskingForName 👀

HIPPIEKICK (approved):
Great job building this chatbot café! ☕

**JavaScript**
- I see you’ve added DOMContentLoaded to ensure that all DOM elements are available before interacting with them. Nice idea!
- Functions like showMessage, handleNameInput, handleDrinkChoice, and others are well-defined and break down the logic into smaller, manageable parts.
- Solid conditions and flow, but it would be nice to see some other input options for users (e.g. buttons), but great that you're validating the input and continue to prompt the user
- There is some repeated logic across the input handling functions (handleDrinkChoice, handleFoodChoice, handleDishChoice). Consider creating a more generic input validation function to simplify the code.
- Great idea clearing the input! You could create a helper function (e.g., clearInput()) to streamline this action.

**Clean Code**
- Be consistent with if you're using single or double quotes

Overall, good job! However I would love to see you go a bit more wild with input options another time 😉

PR #293, opened 2024-09-15
Liselottes chat-bot

IamLise:
## Netlify link
Add your Netlify link here.
PS. Don't forget to add it in your readme as well.

https://splendorous-pegasus-012d17.netlify.app/

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,3 +11,5 @@ Describe how you approached to problem, and what tools and techniques you used t
 ## View it live
 
 Have you deployed your project somewhere? Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+
+https://splendorous-pegasus-012d17.netlify.app/
--- a/code/index.html
+++ b/code/index.html
@@ -23,6 +23,50 @@
             Send
           </button>
         </form>
+        <form id="optionForm">
+          <button id="beverage-btn" class="beverage-btn" type="submit">
+            Beverage
+          </button>
+          <button id="pastry-btn" class="pastry-btn" type="submit">
+            Pastry
+          </button>
+          <button id="cake-btn" class="cake-btn" type="submit">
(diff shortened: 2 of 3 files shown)


HIPPIEKICK (approved):
Hey, nice chatbot! ☕ You've taken a little different approach than we expected (by putting all HTML elements directly in the HTML file), but since you _are_ modifying the styles in JS, we'll allow it 😊 Let's look at some feedback!

**HTML**
- Instead of having a lot of different forms, you could have one form with different sections in it, that you pick up in JS to display/hide

**CSS**
- Since all parts of the form but one, have the display none as default, maybe that could be put in the CSS file? And then you can change whatever you need _when_ you need it in the JS file

**JavaScript**
- The flow of the bot works well and I like that you used buttons ⭐ 
- Your JavaScript code could benefit from additional modularity. For example, if you have repetitive logic (e.g., handling different input types or bot responses), consider splitting these into separate functions for better readability and reusability.
- You're using the onclick method instead of addEventListener, is there a specific reason for this? 👀

**Clean Code**
- Be consistent with whether you're using double or single quotes. If you want to use double quotes, you should change all single quotes from the starter code to be double quotes.
- Remember that functions should be named after what action they are performing
- Great naming of your DOM selectors ⭐

gittebe:
Hej

gittebe (commented):
Hej Liselotte,
The chatbot works well and you used name input and buttons! Well done!!!
that's super interesting how you made it work because you have the forms in your HTML and let them "disappear". I believe if you would have more information/data, it might be easier in terms of shorter and easier to follow to use the innerHTML in the javascript.
Try to make more comments on your code about what you are doing or where one part ends and another starts. That helps me always a lot to read and understand codes.
Overall: Well done!!! :)

